PR #8, opened 2020-06-29
Fix missing pointer dereference

karlosp:
Some tests should be also added for 
```cpp
template<typename Iterable>
    TypeDef& operator+=(const Iterable& children) 
```


mdavidsaver:
Thanks for reporting.  Yes, this needs test coverage.

mdavidsaver:
Added test coverage with 651d7d18fbece66f158477761e58f7d5ca915933

Issue #9, opened 2020-06-30
(De)serialization of Value fields


karlosp:
Is there a way to (de)serialize Value fields, or something similar like the example below in pvData?
```cpp
   epics::pvData::PVStructure::shared_pointer structure_; 
   std::shared_ptr<epics::pvData::ByteBuffer> buffer_;
    ....
    for (size_t i = 0U; i < structure_->getNumberFields(); ++i) {
      //  Request subfield
      const epics::pvData::PVFieldPtr field = structure_->getSubField(i);

      //  If subfield has no children i.e. represents actual value then
      //  serialize it
      if ((field != nullptr) && (field->getNumberFields() == 1)) {
        field->serialize(buffer_.get(), serializable_control_.get());
      }
    }

  // deserialize
  for (size_t i = 0U; i < structure_->getNumberFields(); ++i) {
    const epics::pvData::PVFieldPtr field = structure_->getSubField(i);
    if ((field != nullptr) && (field->getNumberFields() == 1)) {
      field->deserialize(buffer_.get(), deserializable_control_.get());
    }
  }
```

mdavidsaver:
This is not (yet) part of the public API.  Can you say something about your use case?  Is this for file storage, network transport, or something else?  Also, what do you use for `serializable_control_`?  endian-ness?

Have you seen the `serializeToVector()` and related "high level" wrappers?

https://github.com/epics-base/pvDataCPP/blob/79b02254c4f71f5eebb5d27175e5075816a64da4/src/misc/pv/serialize.h#L162-L196

karlosp:
We must replace pvDataCPP lib with this one in a real-time framework that could use EPICS as a transport layer between nodes, processes, or threads.

I took a quick look on `serializeToVector()`, yes probably we could use that, I do not know why it was not used in the first place, but as mentioned all this code will be replaced anyway.

`serializable_control_` is used only for the sake of API. `virtual void Serializable::(de)serializ` takes (De)serializableControl which is pure virtual class. We implemented just two functions:

```cpp
void SerializableControlImpl::alignBuffer(const std::size_t alignment) {
  m_buffer_->align(alignment);
}
void SerializableControlImpl::cachedSerialize(const std::shared_ptr<const epics::pvData::Field>& field,
                                              epics::pvData::ByteBuffer* const buffer) {
  field->serialize(buffer, this);
}
```

Is there some timeline to make (de)serialization part of the public API?

mdavidsaver:
[...]

PR #10, opened 2020-07-06
add user encode/decode

mdavidsaver:
Attempt to address #9 with a user API for (de)serialization.  Three pairs of functions are added for the 3 contexts in which  PVD are found in the PVA protocol: Type, Full structure, and Partial structure.  The `xcode::encode*()` functions append to an `std::vector<uint8_t>`.  The `xcode::decode*()` functions operate on a pair of pointers (like iterators).  The first pointer is passed by reference, and updated based on the number of bytes consumed by each operation.

### DIFF
--- a/src/dataencode.cpp
+++ b/src/dataencode.cpp
@@ -779,6 +779,96 @@ void from_wire_type_value(Buffer& buf, TypeStore& ctxt, Value& val)
         from_wire_full(buf, ctxt, val);
 }
 
-}} // namespace pvxs::impl
+} // namespace impl
+
+namespace xcode {
+
+void encodeType(std::vector<uint8_t>& buf, const Value& prototype, bool be)
+{
+    VectorOutBuf out(be, buf);
+    to_wire(out, Value::Helper::desc(prototype));
+    if(!out.good())
+        throw std::runtime_error("Encode error");
+
+    // trim unused space
+    buf.resize(out.consumed());
+}
+
+void encodeFull(std::vector<uint8_t>& buf, const Value& value, bool be)
+{
+    if(!value)
+        throw std::invalid_argument("Can't encode empty Value");
+
+    VectorOutBuf out(be, buf);
+    to_wire_full(out, value);
+    if(!out.good())
+        throw std::runtime_error("Encode error");
+
+    // trim unused space
+    buf.resize(out.consumed());
+}
+
+void encodeValid(std::vector<uint8_t>& buf, const Value& value, bool be)
+{
+    if(value.type()!=TypeCode::Struct)
+        throw std::invalid_argument("partial/valid encode requires Struct");
+
+    VectorOutBuf out(be, buf);
+    to_wire_valid(out, value);
+    if(!out.good())
+        throw std::runtime_error("Encode error");
+
+    // trim unused space
+    buf.resize(out.consumed());
+}
+
+Value decodeType(const uint8_t*& buf, const uint8_t *bufend, bool be)
+{
+    // buf not actually modified
+    FixedBuf inp(be, const_cast<uint8_t*>(buf), bufend-buf);
+    TypeStore cache;
+    Value ret;
+    from_wire_type(inp, cache, ret);
+    if(!inp.good())
+        throw std::runtime_error("Decode error");
+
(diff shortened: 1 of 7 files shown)


mdavidsaver:
@karlosp Can you comment on whether this change would be flexible enough to support your use case?

karlosp:
I think that we have a bit of special use cases for serialization.

Proposed changes would be great if we would need to save/send `std::vector<uint8_t> buf`.

Because we use the EPICS transport layer in a real-time framework for custom structures, we need to (de)serialize every field separately and not the whole Value.

In the example below I will try to describe how we solve it for now. 
Do not know if this is the best approach but for now, somehow works.

In initialization time we must dynamically construct Value, because we must support custom user types.
For example, let say that we have a struct below we would like to send over EPICS.
e.g.
```cpp
struct TestStructure {
  std::vector<size_t> values;
  size_t index;
  size_t hash;
}
```
Currently, we dynamically create Value and reserve buffer as follows:
```cpp 
pvxs::TypeDef pvxs_field_builder_{pvxs::TypeCode::Struct,"simple_t",{}};
std::vector<pvxs::Member>& pvxs_members;
std::vector<uint8_t> pvxs_buffer_;

// pseudocode
size_t buffer_size{0};
for member in TestStructure:
 pvxs_members.push_back(pvxs::Member(member_type, member_name));
 buffer_size += sizeof(member_type);

pvxs_field_builder_ += pvxs_members_;
Value value_ = pvxs_field_builder_.create();
pvxs_buffer_ = std::vector<uint8_t>(buffer_size, '\0');
```

Example of putting data from TestStructure to PutBuilder.
```cpp
// Sending data
// pseudocode
copy data from TestStructure to pvxs_buffer_;
unsigned char* storage_location = nullptr;
auto put = pvxs::client::PutBuilder(context_.put(topic_name));
size_t position{0};
// pseudocode
for (pvxs::Value fld : value_.iall()) {
  if(fld.type().scalarOf().code == pvxs::TypeCode::Struct)
     continue;
  storage_location = pvxs_buffer_.data() + member_position[position].offset;
  put.set(value_.nameOf(fld), pvxs::shared_array<const int8_t>((int8_t*)(storage_location), sizeof(member) / sizeof(int8_t)));
++position;
}
put.exec()->wait(5.0);
```

receiving data
```cpp
//pvxs::client::Subscription> monitor;
value_.assign(monitor->pop());
// pseudocod
//create empty TestStructure (take special care for initializing vectors and arrays to be correct size)
TestStructure TS;
for (auto fld : value_.iall()) {
  if (fld.type() == pvxs::TypeCode::Struct) {
    continue;
  }
 read and copy data from fld to TS // type end size taking into account
}
```

mdavidsaver:
> Proposed changes would be great if we would need to save/send `std::vector<uint8_t> buf`.

What interface would work?

FYI.  If your libraries are able to interface with a shared_ptr, then the [aliasing constructor](https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr) of `std::shared_ptr` may be helpful here.  eg.

```
auto buf = std::make_shared<std::vector<uint8_t>>({1,2,3});
std::shared_ptr<uint8_t> aliased(buf, buf.data());
buf.reset();
aliased.reset(); // ~vector happens here
```

The aliased shared_ptr shares the same ref. counter, and the same (virtual) destructor.

Also, I'm having difficulty in parsing through your example.  A couple of things I pick up:

> ```c++
>  buffer_size += sizeof(member_type);
> ```

You would like a way to find the serialized size of a field based solely on the type (`TypeCode`)?  The PVD encoding has a number of variable sized elements, which makes this difficult for all but the simplest definitions.

Is the intent to avoid resize()ing the vector during serialization?

> ```c++
> put.set(value_.nameOf(fld), pvxs::shared_array<const int8_t>((int8_t*)(storage_location), sizeof(member) / sizeof(int8_t)));
> ```

Do I understand this correctly?  You would like to send serialized data as a byte array over PVA?

mdavidsaver:
For lack of feedback, I have dropped this change from my development area.

mdavidsaver:
Closing pending feedback (or a clearer use case).

PR #17, opened 2021-04-15
Create extra symlinks for old SONAMES?

mdavidsaver:
For ELF targets, currently only one qualified SONAME is installed `libpvxs.so.0.1`.  This doesn't capture the third (maintenance) version of a build.  So far I've been able to avoid removing/changing symbols, so code linked against older libpvxs can run against a newer release.  But I have added some, so this scheme doesn't fail well if the opposite happens, newer code against older libpvxs.

I've wondering if a better way to handle this is to include the full version in the SONAME, and also create additional symlinks for older compatible releases?

eg. if I had done this so far, then a future 0.1.4 release would install

```
$ ll lib/linux-x86_64/libpvxs.so*
... lib/linux-x86_64/libpvxs.so -> libpvxs.so.0.1.4
... lib/linux-x86_64/libpvxs.so.0.1.0 -> libpvxs.so.0.1.4
... lib/linux-x86_64/libpvxs.so.0.1.1 -> libpvxs.so.0.1.4
... lib/linux-x86_64/libpvxs.so.0.1.2 -> libpvxs.so.0.1.4
... lib/linux-x86_64/libpvxs.so.0.1.3 -> libpvxs.so.0.1.4
... lib/linux-x86_64/libpvxs.so.0.1.4
```

(I think in Mach-O land this would be expressed as `-current_version 0.1.4 -compatibility_version 0.1.0`)

This PR an attempt to add some RULES to do this.  eg. the preceding would be the result of:

```make
COMPAT_VERSIONS = 0.1.0 0.1.1 0.1.2 0.1.3
```

@anjohnson Any thoughts on the ideas or method?

### DIFF
--- a/configure/CONFIG_SITE
+++ b/configure/CONFIG_SITE
@@ -45,3 +45,13 @@ CHECK_RELEASE = YES
 USR_CPPFLAGS_WIN32 += -DNOMINMAX -D_WIN32_WINNT=_WIN32_WINNT_VISTA
 
 USR_CPPFLAGS += -DUSE_TYPED_RSET
+
+# applied to all libraries
+SHRLIB_VERSION = $(PVXS_MAJOR_VERSION).$(PVXS_MINOR_VERSION).$(PVXS_MAINTENANCE_VERSION)
+
+# for ELF targets w/ shared libraries, emit additional symlinks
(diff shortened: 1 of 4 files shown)


mdavidsaver:
Not worth the effort...

PR #19, opened 2021-05-06
bug fix: 'numeric_limits' is not a member of 'std'

basilaljamal:
src/conn.cpp: add limits header to fix 'numeric_limits' is not a member of 'std'

### DIFF
--- a/src/conn.cpp
+++ b/src/conn.cpp
@@ -4,6 +4,8 @@
  * in file LICENSE that is included with this distribution.
  */
 
+#include <limits>
+
 #include <epicsAssert.h>
 
 #include <pvxs/log.h>


mdavidsaver:
Thanks.  In what environment did you see this error?  (OS/compiler versions)

basilaljamal:
Fedora 34
Kernel 5.11.15-300
GCC 11.0.1

mdavidsaver:
Ah, that explains why I didn't see this error.  My current CI testing only goes up to GCC 9.3 ([ubuntu-20.04](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#language-and-runtime)).  I will need to figure out a different recipe for testing older GCC (cf. https://github.com/epics-base/ci-scripts/issues/62).  Solving that will hopefully also provide a way to test newer GCC as well.

PR #22, opened 2021-10-13
Support for multicast and IPv6

mdavidsaver:
Added support for multicast search and beacons over IPv4 and IPV6.

## User configuration

EPICS_PVA_ADDR_LIST and friends accept expanded entries of the form `<IP>[:port][,TTL#][@iface]`.  Where IP may be IPv4 dotted quad or an IPv6 address enclosed in square brackets.  Some examples:

* `192.168.1.1` as before is a unicast address
* `224.0.2.3,255@192.168.1.1` specifies a multicast group `224.0.2.3` reached via the interface with address `192.168.1.1` where UDP packets sent will have TTL of 255.
* `ff02::42:1,1@br0` an IPv6 link scope multicast group `ff02::42:1` reached via the interface `br0`, using a TTL of 1.  (this being a rough functional equivalent to an IPv4 broadcast address)

Note that the `@` syntax is overlapping with the IPV6 specific `#` syntax for specifying interfaces.  Currently this is always resolved in favor of the `@` syntax.

## Sockets

* By default a TCP socket bound to `0.0.0.0` will be "upgraded" to an IPv6 socket bound to `::` allowing incoming connections over either IP version.
* Sockets bound to a specific IPv4 or IPv6 address will of course be limited to that IP version.
* UDP sockets are limited to either IPv4 or IPv6 as it is not possible for an `AF_INET6` socket to interact with IPv4 multicast groups through 4to6 address mapping.

## Current State

At present IPv6 UDP handling must be explicitly enabled.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -546,6 +546,11 @@ def define_DSOS(self):
 
     src_pvxs = [os.path.join('src', src) for src in src_pvxs]
 
+    if OS_CLASS=='WIN32':
+        src_pvxs += ['src/os/WIN32/osdSockExt.cpp']
+    else:
+        src_pvxs += ['src/os/default/osdSockExt.cpp']
+
     event_libs = []
     if OS_CLASS=='WIN32':
         event_libs = ['ws2_32','shell32','advapi32','bcrypt','iphlpapi']
--- a/setup.py
+++ b/setup.py
@@ -542,6 +542,7 @@ def define_DSOS(self):
         'clientintrospect.cpp',
         'clientget.cpp',
         'clientmon.cpp',
+        'clientdiscover.cpp',
     ]
 
     src_pvxs = [os.path.join('src', src) for src in src_pvxs]
--- a/src/Makefile
+++ b/src/Makefile
@@ -106,6 +106,7 @@ LIB_SRCS += clientconn.cpp
 LIB_SRCS += clientintrospect.cpp
 LIB_SRCS += clientget.cpp
 LIB_SRCS += clientmon.cpp
+LIB_SRCS += clientdiscover.cpp
 
 LIB_LIBS += Com
 
--- a/src/Makefile
+++ b/src/Makefile
@@ -86,6 +86,8 @@ LIB_SRCS += nt.cpp
 LIB_SRCS += evhelper.cpp
 LIB_SRCS += udp_collector.cpp
 
+LIB_SRCS += osdSockExt.cpp
+
 LIB_SRCS += config.cpp
 LIB_SRCS += conn.cpp
 
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -468,8 +468,7 @@ Value buildCAMethod()
 }
 
 ContextImpl::ContextImpl(const Config& conf, const evbase& tcp_loop)
-    :canIPv6(evsocket::canIPv6())
-    ,ifmap(IfaceMap::instance())
+    :ifmap(IfaceMap::instance())
     ,effective(conf)
     ,caMethod(buildCAMethod())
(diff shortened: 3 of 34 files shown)


mdavidsaver:
Some example configurations I'm testing.

All IPv4 multicast

```sh
PVXS_LOG=*=DEBUG \
EPICS_PVA_AUTO_ADDR_LIST=NO \
EPICS_PVA_ADDR_LIST=224.0.1.1,1@127.0.0.1 \
EPICS_PVAS_INTF_ADDR_LIST='0.0.0.0 224.0.1.1,1@127.0.0.1' \
./example/O.linux-x86_64-debug/mailbox box

PVXS_LOG=*=DEBUG \
EPICS_PVA_AUTO_ADDR_LIST=NO \
EPICS_PVA_ADDR_LIST=224.0.1.1,1@127.0.0.1 \
./bin/linux-x86_64-debug/pvxget box
```

All IPv6 multicast

```sh
PVXS_LOG=*=DEBUG \
EPICS_PVA_AUTO_ADDR_LIST=NO \
EPICS_PVA_ADDR_LIST=ff02::42:1,1@br0 \
EPICS_PVAS_INTF_ADDR_LIST='0.0.0.0 ff02::42:1,1@br0' \
./example/O.linux-x86_64-debug/mailbox box

PVXS_LOG=*=DEBUG \
EPICS_PVA_AUTO_ADDR_LIST=NO \
EPICS_PVA_ADDR_LIST=ff02::42:1,1@br0 \
./bin/linux-x86_64-debug/pvxget box
```

mdavidsaver:
> There is a problem, however, at least on Mac OS ...

Ah Kay, you've lifted the scab on another significant OS dependent behavior...  I'm working on some test coverage to probe the behavior of `bind()`ing sockets to both `0.0.0.0` and `::` with the same port.  What I see so far is:

* Linux doesn't allow this for TCP (w/o `IPV6_V6ONLY`), or UDP (w/o `REUSEADDR/PORT`).
* Windows and RTEMS5+libbsd allow both to be bound in either order.
* OSX allows binding `0.0.0.0` and then `::`, but errors on the second if the order is reversed.

I'm also surprised to see OSX and RTEMS/libbsd behaving differently.  I'm under the impression that both are ~recent forks of a *BSD network stack.

So it looks like avoiding conflicts w/ IPv4 implementations is going to be different for each target.

I'm going to try the following as a starting point for TCP listeners:

* Linux.  Need to magically "promote" `0.0.0.0` to `::` since there can be only one.
* Windows/RTEMS/OSX.  Need to bind both, starting with `0.0.0.0` to detect a conflict.

I'm also going to have to add a test case which tests binding to the wildcard address(es).  Currently all high level testing uses a configuration w/ `127.0.0.1`.

(fun fact, Docker doesn't configure IPv6.  So container builds can't run tests of IPv6)

mdavidsaver:
@kasemir I think I've addressed the `bind()` inter-op issue w/ OSX.  Can you re-test?

kasemir:
Yes, now good.

To test, I'm starting a softIocPVA which binds to tcp4 *.5075, then starting the pvxs mailbox example.

Before, mailbox would bind to tcp46 *.5075, not noticing that port is already bound as tcp4.
Now, starting the mailbox example detects "Server unable to bind port 5075, falling back to 0.0.0.0:52648".

kasemir:
Another improvement: The build tended to fail in one of the later test code stages. Now all builds without errors on Mac OS X.

mdavidsaver:
Updated to fix an issue with automatic configuration which was preventing beacons from being sent by default on Linux.

PR #26, opened 2022-03-25
src/conn.cpp:  add limits header to fix error: 'numeric_limits' is not a member of 'std' 

basilaljamal:
src/conn.cpp:  add limits header to fix error: 'numeric_limits' is not a member of 'std'


mdavidsaver:
I'm not sure if you intended to close this PR.  I see that `<limits>` isn't included anywhere explicitly, which is a real omission.  I've pushed fe6974025ab0e89ad60a41a7d0fdce181f544da8 which adds this include to the `src/utilpvt.h` internal header, where it will apply to usages in both `src/conn.cpp` and `src/config.cpp`.

PR #35, opened 2023-03-10
Meson build

minijackson:
As discussed with @mdavidsaver during the 2023 Codeathon.

Currently uses sources from epics-base/epics-base#358 and epics-base/ci-scripts#77, so keeping this PR as draft until they are merged.

This is kind of a long-shot PR, in that it allows building PVXS without going through epics-base Makefile framework. This would allow easier integration into other build / packaging systems (building Python wheels, CI, Debian packaging, etc.).

For examples, I have tested cross-compilation with Nix, and it works without any special care.

Right now, this needs an already compiled epics-base, and uses the pkg-config file that epics-base produces.

By default, the output follows the same hierarchy as epics-base (`bin/linux-x86_64`), but it can be configured not to, by specifying `-Depics_install_hierarchy=false` to the `meson` invocation.


src/meson.build
```diff
@@ -0,0 +1,126 @@
...
+    'mkdir @0@/pvxs; cp @1@/@2@ @1@/@3@ @1@/@4@ @0@/pvxs/'.format(
+      meson.current_build_dir(),
+      fs.parent(meson.current_build_dir()),
+      version_num_h,
+      vcs_h,
+      describe_h,
+    ),
+  ],
+  check: true,
+)
```
ericonr, line 43: Is this done

tools/meson.build
```diff
@@ -0,0 +1,19 @@
...
+    ['pvxinfo', ['./info.cpp']],
+    ['pvxmonitor', ['./monitor.cpp']],
+    ['pvxput', ['./put.cpp']],
+    ['pvxcall', ['./call.cpp']],
+    ['pvxlist', ['./list.cpp']],
+    ['pvxmshim', ['./mshim.cpp']],
+  ]
+  executable(
+    e[0],
+    e[1],
```
ericonr, line 12: You're using `foreach e` here, in tools, and `foreach t` in examples. Aren't they switched around?

src/meson.build
```diff
@@ -0,0 +1,126 @@
...
+run_command(
+  [
+    'sh',
+    '-c',
+    'mkdir @0@/pvxs; cp @1@/@2@ @1@/@3@ @1@/@4@ @0@/pvxs/'.format(
+      meson.current_build_dir(),
+      fs.parent(meson.current_build_dir()),
+      version_num_h,
+      vcs_h,
+      describe_h,
```
ericonr, line 39: As far as I can tell, the EPICS build does not install `pvxsVCS.h` or `describe.h`.

tools/meson.build
minijackson: I'm using `e` for executable here, and `t` for tests elsewhere. I should probably change to `e` in `examples`, since they're not really "tests"

src/meson.build
minijackson: You're right! I'll remove them from the `install_headers` below

tools/meson.build
ericonr: Sounds good to me!

PR #52, opened 2023-08-07
fix pvxlist help message

ericonr:
piping into pvxlist doesn't work, but passing the servers as argument to it does

### DIFF
--- a/tools/list.cpp
+++ b/tools/list.cpp
@@ -43,7 +43,7 @@ void usage(const char* argv0)
             "   "<<argv0<<" -w 0 -v\n"
             "\n"
             "  List all PV names.  (Warning: high network load)\n"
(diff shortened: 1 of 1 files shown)


tools/list.cpp
```diff
@@ -43,7 +43,7 @@ void usage(const char* argv0)
             "   "<<argv0<<" -w 0 -v\n"
             "\n"
             "  List all PV names.  (Warning: high network load)\n"
-            "   "<<argv0<<" -w 5 | "<<argv0<<"\n"
+            "   "<<argv0<<" $("<<argv0<<" -w 5)\n"
```
mdavidsaver, line 46: Good point.  My original intent was `pvxlist ... | xargs pvxlist`, which I think would have the same effect as `pvxlist $(pvxlist ...)`.

PR #53, opened 2023-08-17
TLS

mdavidsaver:
Applying OpenSSL to PVXS.

Wire compatibility with existing (plain tcp) clients/servers is maintained.  Likewise this PR is compatible with work by @kasemir adding TLS support to [core.pva](https://github.com/ControlSystemStudio/phoebus/blob/master/core/pva/TLS.md) in the phoebus repository.

To maintain compatibility.  When configured with a keychain file, client search requests include two "protocol" names: `tcp` and `tls`.  Server listens on a second TCP port for TLS connections, prefers to respond with `tls` when both present.

Adds `x509` AUTHZ method.  Client advertises `x509` when configured with a client certificate.  Server prefers `x509` to `ca`.  If selected, uses client cert `commonName` as account name.  eg. `CN=foo` appears to ACF logic as `foo`.  (note, this really needs an extension to the ACF logic to account for different source of accounts)

TLS features/restrictions

- OpenSSL >= 3.0
- TLS protocol version >= 1.3 required
- ALPN extension negotiates with protocol `pva/1`.  (mandatory?)
- Client/server reload TLS configuration (eg. replace cert) at runtime.  Forces closes connections.

New Configuration:

- `$EPICS_PVAS_TLS_KEYCHAIN` `$EPICS_PVA_TLS_KEYCHAIN` name of a PKCS12 file containing some combination of key, server/client certificate, and/or CA certificates.  The file path may optionally be followed by `;` and a password.  eg. `EPICS_PVAS_TLS_KEYCHAIN=/path/to/some.p12;secret`.
- `$EPICS_PVAS_TLS_OPTIONS` `$EPICS_PVA_TLS_OPTIONS` space separated list of key=value pairs.  Currently supported are `client_cert=optional` and `client_cert=require`.
- `$EPICS_PVAS_TLS_PORT` `$EPICS_PVA_TLS_PORT` default port to listen for TLS connections
- `$OPENSSL_CONF` Location of OpenSSL [config file](https://www.openssl.org/docs/man3.1/man5/config.html).  Defaults to `/usr/lib/ssl/openssl.cnf` on Linux.  Application name `pvxs` is used.
- `$SSLKEYLOGFILE` If set, TLS session keys will be written here. eg. for use by [wireshark](https://wiki.wireshark.org/TLS).  cf. https://github.com/mdavidsaver/cashark/pull/14  (may be disabled at build time with `make PVXS_ENABLE_SSLKEYLOGFILE=NO ...`)

Building:

`libevent` must be built with optional openssl support (bundled build will detect).  `openssl` is expected to be installed in the default search path.  I have no plans to bundled openssl.

TODO:

- [X] Configure TLS in `$EPICS_PVA_NAME_SERVERS` w/ URI-ish `pvas://1.2.3.4:5678`.
- [X] Verify interoperability with PKCS12 created by java `keytool`  (OpenSSL parser for PKCS12 is... limited)
  - [X] Can work with files created by `openssl pkcs12` or `keytool`.  (although Java can't current work with openssl created files)
  - [X] Handle password protected PKCS12 (which java keytool insists on...)
- [X] Configure client cert. required at TLS level?
- [X] Fix OSX RPATH issue
- [x] Fix MSVC build failure w/ appveyor
- [X] IOC shell commands to manage TLS (re)configuration
- [X] Test server reconfigure
  - [X] Expose server credentials to client
- [X] `x509` AUTHZ method design
  - [X] Add `authority` alongside `method` and `account`
- [X] Update python module build
- [ ] Config for openssl outside of implicit search path (OSX brew).
- [ ] What to do about CMD_BEACON ?
- [ ] OSCP over PVA?
- [ ] What to do about certificate expiration?  (drop credentials?  force reconnect w/o TLS?)
- [ ] Include CA IDs in search request/reply?  (more graceful fallback if mis-matched CA certs)
- [ ] Investigate openssl provider/engine support.  eg [tpm2-openssl](https://github.com/tpm2-software/tpm2-openssl)

Supersedes https://github.com/mdavidsaver/pvxs-dev/pull/2


mdavidsaver:
I think this work has progressed far enough to be (somewhat) more widely testable.  The unittest process creates a consistent set of TLS certificates which may be useful for some manual testing.  Though I would not recommend them for anything beyond some basic exploration.  Some knowledge of x509 certificates will be needed to go beyond the recipe shown below.

An example of running the `ticker` example server, and subscribing with `pvxmonitor`.

```
EPICS_PVA_TLS_KEYCHAIN=test/O.linux-x86_64/server1.p12 \
  ./example/O.linux-x86_64/ticker cnt
```

```
EPICS_PVA_TLS_KEYCHAIN=test/O.linux-x86_64/ca.p12 \
  ./bin/linux-x86_64-debug/pvxmonitor cnt
```

Verification is easier with an IOC.

```
cat <<EOF > tick.db
record(calc, "\$(P=)cnt") {
  field(INPA, "\$(P=)cnt")
  field(CALC, "A+1")
  field(SCAN, "1 second")
}
EOF
EPICS_PVA_TLS_KEYCHAIN=test/O.linux-x86_64/server1.p12
  ./bin/linux-x86_64/softIocPVX -d tick.db
INFO: PVXS QSRV2 is loaded and ENABLED.
Starting iocInit
...
epics> dbl
cnt
epics> pvxsr 3
...
TLS Cert. subject:CN=server1 issuer:CN=intermediateCA from: Aug 17 17:00:23 2023 GMT until: Aug 14 17:00:24 2033 GMT
    Peer[fe80::7816:e3ff:fe25:5208]%5:59152 backlog=0 TX=994 RX=120 auth=ca TLS
        Cred: ca/mdavidsaver@[fe80::7816:e3ff:fe25:5208]%5:59152
        cnt TX=927 RX=46         Executing ioid=268443648 MONITOR
epics>
```

On the client side, switching from `ca.p12` to `client1.p12` will use client certificate authentication.  Then `Cred: ca/mdavidsaver...` becomes `Cred: x509rootCA/client1...`.

```
...
TLS Cert. subject:CN=server1 issuer:CN=intermediateCA from: Aug 17 17:00:23 2023 GMT until: Aug 14 17:00:24 2033 GMT
    Peer[fe80::7816:e3ff:fe25:5208]%5:40564 backlog=0 TX=1035 RX=86 auth=x509 TLS
        Cred: x509rootCA/client1@[fe80::7816:e3ff:fe25:5208]%5:40564
        Cert: subject:CN=client1 issuer:CN=intermediateCA from: Aug 17 17:00:24 2023 GMT until: Aug 14 17:00:25 2033 GMT
        cnt TX=968 RX=46         Executing ioid=268443648 MONITOR
```

mdavidsaver:
Update.  @kasemir  When built against what will become openssl 3.2.x, `test/gen_test_certs.cpp` will now produce `.p12` files which work with java.  (see `#ifdef NID_oracle_jdk_trustedkeyusage`)  Note that this does not yet apply to files produced by `openssl pkcs12 ...`.

mdavidsaver:
Updated to add `$EPICS_PVAS_TLS_OPTIONS` `$EPICS_PVA_TLS_OPTIONS` as space separated list of key=value pairs. Currently supported are `client_cert=optional` (default) and `client_cert=required`.  Where `required` causes a server to set `SSL_VERIFY_FAIL_IF_NO_PEER_CERT` so that the TLS handshake will fail unless a valid client cert is presented.

mdavidsaver:
I have also added another make variable `OPENSSL` which should function like `LIBEVENT` to set an external/non-epics installation prefix path when these packages are outside of the default search paths.

Also like `LIBEVENT`, `OPENSSL` should only be set during PVXS builds.  The result is captured and written to `cfg/TOOLCHAIN_PVXS.$(T_A)` for use by dependent modules.

So if all goes well, the added dependency on libssl should not require changes to downstream Makefiles...

mdavidsaver:
Rebased branch, and extended the output of `pvxinfo -v` to show server credentials, which includes TLS vs. plain PVA.

```
./bin/linux-x86_64/pvxinfo cnt -v
...
# anonymous/@[fe80::9453:55ff:fe80:63a1]%4:5075
cnt from [fe80::9453:55ff:fe80:63a1]%4:5075
...
```

The server is "anonymous" as plain PVA does not send server credentials.

```
EPICS_PVA_TLS_KEYCHAIN=test/O.linux-x86_64/client1.p12 \
 ./bin/linux-x86_64/pvxinfo cnt -v
...
# TLS x509:rootCA/ioc1@[fe80::9453:55ff:fe80:63a1]%4:5076
cnt from [fe80::9453:55ff:fe80:63a1]%4:5076
...
```

Here the server is "ioc1" through use of `EPICS_PVA_TLS_KEYCHAIN=test/O.linux-x86_64/ioc1.p12`.

mdavidsaver:
Found it.  This PR, and some others, have somehow fallen out of the PR list.  Maybe a comment will bring it back?

edit: Yes, now this PR appears in the `/pulls` list again.  Weird...

PR #58, opened 2023-09-13
Enable PVXS_QSRV_ENABLE by default

tboegi:
Switch the default value for PVXS_QSRV_ENABLE from "NO" to "YES". This means that "qsrv" is enabled, once compiled and linked into the IOC. It can be switched off in runtime:

epicsEnvSet("PVXS_QSRV_ENABLE", "NO")

before calling iocInit()

### DIFF
--- a/documentation/ioc.rst
+++ b/documentation/ioc.rst
@@ -48,9 +48,9 @@ Beginning with PVXS 1.2.0 the functionality of `QSRV <https://epics-base.github.
 is replicated in the ``pvxsIoc`` library.
 Currently this feature preview is considered **alpha** level, with missing functionality.
 In addition to linking in ``libpvxsIoc``,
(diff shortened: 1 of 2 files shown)


mdavidsaver:
eddc6870217de329e8eafcb268f8570f76258deb includes a mostly equivalent change.  Going forward, QSRV2 will be enabled by default unless QSRV1 is also loaded.  This default can be overridden by setting `$PVXS_QSRV_ENABLE` or `$EPICS_IOC_IGNORE_SERVERS`.

QSRV1 is detected by checking if a device support named `devWfPDBDemo` is registered (from `qsrv.dbd`).  The functional equivalent in QSRV2 is `devWfPDBQ2Demo` as the "DEMO" device supports must have different names.

PR #63, opened 2023-11-21
JSON parsing, applied to pvxput

mdavidsaver:
Wrap YAJL parser from Base, which **requires Base >=3.15**.  Teach `pvxput` to make use of this.

Adds special case handling of `enum_t` at a low level through the `Struct` node.

Relaxes requirement to prefix Union member access with `->` at the beginning of an expression.

TODO:

- [ ] Assign `Any` by inferring a `TypeDef`.
- [ ] Handle uint64

### DIFF
--- a/coverage.sh
+++ b/coverage.sh
@@ -45,7 +45,7 @@ OUTDIR="$PWD"/coverage
 install -d "$OUTDIR"
 
 # pre-process for each directory, correcting "file" relative to repo root
-for dir in src ioc
+for dir in src ioc tools
 do
     ( cd "$TDIR"/$dir/O.linux-* \
     && gcovr --gcov-ignore-parse-errors -v -r .. --json \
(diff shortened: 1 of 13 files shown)


mdavidsaver:
@klemenv wrt. #3 are you still using PVXS with Base 3.14?

mdavidsaver:
My recent addition of ABI diff to GHA (de12bc181e065d7120074d351157b08f89543e0c) turns up the following, which I can replicate locally.  I am not sure what to make of it.  Why would this non-inline function apparently not follow the Linux/amd64 [calling convention](https://en.wikipedia.org/wiki/X86_calling_conventions)?

image

mdavidsaver:
Applied as 5ab17ec375e93ccaa75c00a413fb9d64a5473e26 as an Expert API.

PR #64, opened 2023-11-24
Documentation for PVA links

mdavidsaver:
Adds documentation of PVA links.

Partly derived from the [original](https://github.com/epics-base/pva2pva/blob/master/documentation/qsrvpage.dox) with some new material.


mdavidsaver:
@simon-ess I would appreciate any thoughts or changes which might occur to you.

PR #67, opened 2024-04-30
fix: use prepare cleanup hooks when epics-base>=7.0.8.0

minijackson:
Found while trying to update my meson PR, it seems this feature was added in 7.0.8.0: https://github.com/epics-base/epics-base/commit/9f660f2238a6efe85af8a49510edeaee69b6641a, but the logic uses this feature for 7.0.7.1


mdavidsaver:
> this feature was added in 7.0.8.0

I think the correct logic would be `>= 7.0.8`.

```
EPICS_VERSION_INT >= VERSION_INT(7, 0, 8, 0)
```

> but the logic uses this feature for 7.0.7.1

oops...  Clearly my mistake.

PR #76, opened 2024-08-01
Release 1.4.0a1

OCopping:


mdavidsaver:
As it stands, all changes since `1.3.1` maintain ABI compatibility.  So the next release can be `1.3.2` (and so `1.3.2a1` the first pre-release).

fyi. assessing ABI changes can be difficult.  A first pass check is to look for changes to installed/public headers.  In this case, only non-functional changes to URLs.

```
$ git diff --stat 1.3.1..origin/master -- src/pvxs ioc/pvxs
 src/pvxs/version.h | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)
```

As an automated check, run `./abi-diff.sh 1.3.1 HEAD` which wraps `abi-dumper` and `abi-compliance-checker`.  Currently two separate reports are produced for the two installed libraries `pvxs` and `pvxsIoc`.  This script currently assumes the a Linux host.

```
$ ./abi-diff.sh 1.3.1 e05ee2b2abd7606e548549b094c62786ea082a20
...
Binary compatibility: 100%
Source compatibility: 100%
...
Binary compatibility: 100%
Source compatibility: 100%
...
```

Beyond this things get difficult, and become a question of judging whether the semantics of a function changes enough to be considered incompatible.

Issue #81, opened 2024-08-14
evconnlistener_new() fails


mdavidsaver:
Observed on a Debian 12 host with an IOC, one of 34 PVA servers started in parallel during system boot.  A later restart succeeds.

```
...: Error in pvxsBaseRegistrar : bad_alloc serverconn.cpp:448
```

https://github.com/epics-base/pvxs/blob/5fa743d4c87377859953012af3c0fbcd1b063129/src/serverconn.cpp#L447-L449

`evconnlistener_new()` can fail from `malloc()` (seems unlikely) or `listen()`.  The later seems the most likely.  By this point, the socket is already bound, which I thought was the only reason `listen()` could fail (aside from alloc...).

Maybe `bind()` on Linux is lazier than I thought?

`ev_owned_ptr<>` should be changed to capture `SOCKERRNO`.

mdavidsaver:
Well, surprise.  `listen()` for ipv4 on Linux is lazier than I expected.  And there is a [confirmed](https://lore.kernel.org/netdev/4887534b-cac3-65b5-8c9f-f736c1e56ba3@virtuozzo.com/T/) and documented race condition...

from `inet_csk_listen_start()`

```c
	/* There is race window here: we announce ourselves listening,
	 * but this transition is still not validated by get_port().
	 * It is OK, because this socket enters to hash table only
	 * after validation is complete.
	 */
```

https://github.com/torvalds/linux/blob/9d5906799f7d89c9e12f6d2e0fccb00713c945ab/net/ipv4/inet_connection_sock.c#L1343-L1347

mdavidsaver:
In `ServIface::ServIface()` this could be handled by moving the `evconnlistener_new()` into the TCP port selection loop.  Hopefully I can figure out a way to trigger this situation as some details need further exploration.  eg. according to [this post](https://lore.kernel.org/netdev/4887534b-cac3-65b5-8c9f-f736c1e56ba3@virtuozzo.com/T/) both racers can fail.

mdavidsaver:
Now I think I understand why RSRV tests both `bind()` and `listen()` at once...

```c
            if(bind(tcpsock, &scratch.sa, sizeof(scratch))==0 && listen(tcpsock, 20)==0) {
```

https://github.com/epics-base/epics-base/blob/86cdfc596f402657401b07c3a38493159022d023/modules/database/src/ioc/rsrv/caservertask.c#L195

mdavidsaver:
[...]

PR #82, opened 2024-08-15
server: correctly adjudicate collision bind() of specific port

mdavidsaver:
Attempts to address #81.

On Linux (at least) SO_REUSEADDR, which allows a new listener to bind while an existing sock is in FIN-WAIT.  Apparently this allows any number of sockets to bind(), but only when listen() to succeed.

Further, on Linux there is a known documented race condition which can result in all listen() failing.  It isn't clear how to handle this case without a potentially infinite loop, so ignore it. If this happens, then eg. no PVA server will get port 5075.

So when probing for another listener, it is necessary to enter the listening state.  When this fails, the socket is no longer usable for another bind(), so it is necessary to allocate another for the next attempt.

### DIFF
--- a/src/serverconn.cpp
+++ b/src/serverconn.cpp
@@ -408,18 +408,40 @@ ServIface::ServIface(const SockAddr &addr, server::Server::Pvt *server, bool fal
     server->acceptor_loop.assertInLoop();
     auto orig_port = bind_addr.port();
 
-    sock = evsocket(bind_addr.family(), SOCK_STREAM, 0);
-
-    if(evutil_make_listen_socket_reuseable(sock.sock))
-        log_warn_printf(connsetup, "Unable to make socket reusable%s", "\n");
-
     // try to bind to requested port, then fallback to a random port
     while(true) {
+
+        sock = evsocket(bind_addr.family(), SOCK_STREAM, 0);
+
+        if(evutil_make_listen_socket_reuseable(sock.sock))
+            log_warn_printf(connsetup, "Unable to make socket reusable%s", "\n");
+
         try {
             sock.bind(bind_addr);
+            // semantics of SO_REUSEADDR on *nix allow multiple bind() to success.
+            // however, only one listen() will succeed
+
+            // added in libevent 2.1.1
+#ifndef LEV_OPT_DISABLED
+#  define LEV_OPT_DISABLED 0
+#endif
+
+            const int backlog = 4;
+            auto list(evconnlistener_new(server->acceptor_loop.base, onConnS, this, LEV_OPT_DISABLED|LEV_OPT_CLOSE_ON_EXEC, backlog, sock.sock));
+            if(!list) {
+                int err = evutil_socket_geterror(sock);
+                throw std::system_error(err, std::system_category());
+            }
+            listener = evlisten(__FILE__, __LINE__, list);
+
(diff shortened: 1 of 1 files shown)


mdavidsaver:
The OSX build CI failures will be resolved by https://github.com/epics-base/setuptools_dso/pull/35

anjohnson:
It looks like this is only for stream sockets and TCP, so if no server gets port 5075 that won't prevent UDP searche packets from being received and distributed to other servers via the localhost loopback. Is that correct?

Have you given any thought to how much work might be needed for a server to accept its sockets from inetd (stdin/stdout) or systemd via `sd_listen_fds()`? That might be a useful option for embedded servers, although socket activation seems less likely to make sense for IOCs at least.

mdavidsaver:
> It looks like this is only for stream sockets and TCP ...

Correct.  As I understand it, this laziness of `bind()` is specific to a TCP sockets where the REUSEADDR is set (as so specific to *nix).  eg.

```py

S1=socket(AF_INET, SOCK_STREAM)
S2=socket(AF_INET, SOCK_STREAM)
S1.bind(('127.0.0.1', 5000))
S2.bind(('127.0.0.1', 5000)) # fails!  (EADDRINUSE)
```
```py
S1=socket(AF_INET, SOCK_STREAM)
S2=socket(AF_INET, SOCK_STREAM)
S1.setsockopt(SOL_SOCKET, SO_REUSEADDR, 1)
S2.setsockopt(SOL_SOCKET, SO_REUSEADDR, 1)
S1.bind(('127.0.0.1', 5000))
S2.bind(('127.0.0.1', 5000)) # succeeds!
S1.listen(4)
S2.listen(4) # fails! (EADDRINUSE)
```

mdavidsaver:
> Have you given any thought to ... `sd_listen_fds()`?

No, not really.  It seems like a lot of work for not much benefit, with a high probability of a mis-configured .socket files (plural!) causing chaos.

What I have thought about is calling `sd_notify()` with IOC lifecycle changes.  Primarily using `initHookAfterIocRunning` to emit `READY=1`, so that dependent units don't race CA/PVA server startup.

mdavidsaver:
... of course usage of `sd_*` is mote so long as procServ is involved.

mdavidsaver:
fyi. my attempt at provoking this race was not successful.  I guess a shell loop is too slow with so many `fork()`s.

```sh
cat > tick.db << EOF
record(calc, "$(P=)cnt") {
  field(INPA, "$(P=)cnt")
  field(CALC, "A+1")
  field(SCAN, "1 second")
}
EOF

for n in `seq 1 100`; do sh -c "softIocPVX -m P=$n: -d tick.db -S </dev/null &" ; done
```

Followed by

```sh
for n in `seq 1 100`; do echo $n:cnt; done | xargs pvxget
```

Will complete without timeout if all PVA servers started.

cleanup

```sh
killall softIocPVX
```

mdavidsaver:
> fyi. my attempt at provoking this race was not successful ...

It can, sometimes, eventually.  Looping through [iocBomb.sh](https://github.com/user-attachments/files/16631947/iocBomb.txt) gets a timeout on one or two PVs within a couple of minutes on my laptop without this PR.  With this PR applied, I eventually got bored.

```sh
while sh iocBomb.sh; do date; done
```

I am satisfied with this result.

PR #86, opened 2024-10-21
doc: client fix.

ericonr:
Despite what the raw line wrapping makes it seem, there isn't a line break in the final generated documentation, so the text just goes directly from the environment variable name to its default value.


mdavidsaver:
Good eye.  Thanks.

Issue #88, opened 2024-11-18
QSRV groups aren't created when loaded with `dbLoadGroups` in .cmd file


ericonr:
**Describe the bug**

Defining a group in a .json file to be loaded by `dbLoadGroups` doesn't work.

**To Reproduce**

1. unpack the attached tarball
2. run `softIocPVX st.cmd`
3. run `pvxgl` in iocsh. Only `table-info` is listed

The bug doesn't manifest itself if using `softIocPVX` command line options, though:

1. run `softIocPVX -d records.db -G quote.json`
2. run `pvxgl` in iocsh. Now `table-info` and `table-quote` are listed

**Expected behavior**

`dbLoadGroup` should properly load a group.

**Information (please complete the following):**

```
Host: linux-x86_64
Target: linux-x86_64 Linux gcc
Toolchain
    __cplusplus = 201103
    GCC 11.4.0
    _GLIBCXX_USE_CXX11_ABI = 1
    GLIBC 2.35
    __GLIBCXX__ 20230528
Versions
    PVXS 1.3.2 (1.3.2-6-g8bf16de8a222d9d5fca3-dirty)
    EPICS 7.0.8.2-DEV
    libevent 2.1.12-stable
Runtime
    uname() -> Linux S-SWC03-L 6.8.0-45-generic #45~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Sep 11 15:25:05 UTC 2 x86_64
    epicsThreadGetCPUs() -> 8
```

**Additional context**:

The docs should include an explanation about the requirement to quote `+` prefixed mappings.

ericonr:
[qgroup.tar.gz](https://github.com/user-attachments/files/17806188/qgroup.tar.gz)

ericonr:
So the difference between using `-G` and `dbLoadGroups()` in a cmd file:

`-G` simply calls `pvxs::ioc::dbLoadGroups`

`dbLoadGroups()` calls `pvxs::ioc::dbLoadGroups` then it calls `GroupConfigProcessor().loadConfigFiles();`. Simply removing the latter call makes it so the groups defined in the json file show up in `pvxgl` output.

ericonr:
Fixed in https://github.com/epics-base/pvxs/commit/7e031a20ff8b92954af8e5fc8167ae73fc3eff7b

PR #89, opened 2024-11-19
Small changes to qgroup docs

ericonr:
Partially addresses #88 , in that it mentions the necessity of quoting mapping names.


mdavidsaver:
While a little more verbose, I agree that it makes sense for documentation to show the "universal" syntax.  Thanks.

Issue #97, opened 2025-01-30
DBE_ARCHIVE not handled correctly


mdavidsaver:
QSRV2 tries to handle `record._options.DBE` from a pvRequest to compute the DBE mask passed through to dbEvent.  However, currently `DBE_ARCHVIE` is being lost since the `Everything` mask does not in fact contain every mask.  So subscribing with `record[DBE=ARCHIVE]` will only see property events (since DBE_PROPERTY is handled specially).

https://github.com/epics-base/pvxs/blob/ff1d6510cbf84229f7018eacf4008ecd85dce565/ioc/iocsource.h#L40

https://github.com/epics-base/pvxs/blob/ff1d6510cbf84229f7018eacf4008ecd85dce565/ioc/singlesource.cpp#L82

PR #98, opened 2025-02-01
ioc: fix DBE_ARCHIVE handling w/ singlesource

mdavidsaver:
Stop ignoring DBE_ARCHIVE (#97) for single PV.  Does not address group PV.

~~Add special 'record._options.DBE' to subscription Value to indicate actual mask used.~~  ...Dropped.

### DIFF
--- a/documentation/ioc.rst
+++ b/documentation/ioc.rst
@@ -128,6 +128,22 @@ Currently supported format hints are:
 - Exponential
 - Engineering
 
+When subscribing, the client provided ``pvRequest`` Value may contain the field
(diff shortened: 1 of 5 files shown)


mdavidsaver:
> Add special 'record._options.DBE' to subscription Value to indicate actual mask used.

I am reconsidering adding the field in favor of reporting parse errors via #99.

mdavidsaver:
I have backed out the addition of `record._options.DBE` into subscription updates, which makes the functional part of this PR quite small.  I'm going to include it in PVXS 1.3.3.

PR #100, opened 2025-02-13
Fix MS propagation of AMSG

mdavidsaver:
Call `recGblSetSevrMsg()` when handling `MSI` or `MS` link modifiers.  Also when disconnected.


mdavidsaver:
Applied as of e32a752d12b9cc417b7337ff178dd54d36b4a8f3

PR #101, opened 2025-02-17
Sendto pktinfo

mdavidsaver:
Attempts to address this situation encountered with p4p.gw https://github.com/epics-base/p4p/issues/162 where a PVA client host is intentionally reachable by routes from multiple interfaces of a gateway host.  Because PVXS binds UDP sockets to `0.0.0.0` and `[::]`, a SEARCH reply will currently be sent by the most direct route, which may not be the one by which the original SEARCH request arrived.

This PR changes from `sendto()` to `sendmsg()` with a `IP_PKTINFO` or `IPV6_PKTINFO` control message specifying the requested outbound interface and/or source address.

The source address for a message forwarded by local multicast is taken from the `ORIGIN_TAG` message, when available and when an interface address.  An interface index is looked up from this.

Currently neither override is specified for non-unicast SEARCH replies, nor for BEACON.

This PR also changes the criteria for identifying a UDP message which has been forwarded by a local peer, even when a ORIGIN_TAG is not present.  This is done by looking at OS provided interface and destination address, which I think can be trusted.  This should also make it possible to distinguish traffic arriving over the loopback which has not be forwarded.

wrt. portability.  All major IP stacks I looked at (BSD, Linux, Windows) provide `IPV6_PKTINFO`.  `IP_PKTINFO` is provided by Linux and Windows, but not BSD.  So for RTEMS or OSX, a call to `sendmsg()` will be equivalent to `sendto()`.

@kasemir @FreddieAkeroyd  fyi.

### DIFF
--- a/src/udp_collector.cpp
+++ b/src/udp_collector.cpp
@@ -76,8 +76,8 @@ struct UDPCollector final : public UDPManager::Search,
             if(!(ev&EV_READ))
                 return;
 
-            // handle up to 4 packets before going back to the reactor
-            for(unsigned i=0; i<4 && self->handle_one(); i++) {}
+            // bound number of packets handled before going back to the reactor
+            for(unsigned i=0; i<16 && self->handle_one(); i++) {}
 
         }catch(std::exception& e) {
             log_crit_printf(logio, "Ignoring unhandled exception in UDPManager::handle(): %s\n", e.what());
--- a/src/udp_collector.cpp
+++ b/src/udp_collector.cpp
@@ -96,7 +96,7 @@ struct UDPManager::Pvt {
     SockAttach attach;
 
     evbase loop;
-    IfaceMap& ifmap;
+    IfaceMap ifmap;
 
     // only manipulate from loop worker thread
     // key'd by address family and port#
--- a/src/udp_collector.cpp
+++ b/src/udp_collector.cpp
@@ -58,33 +58,20 @@ struct UDPCollector final : public UDPManager::Search,
     void addListener(UDPListener *l);
     void delListener(UDPListener *l);
 
-    bool handle_one();
+    bool handle_one(const IfaceMap::Current &ifinfo);
 
     enum origin_t {
-        Remote,    // non-local sender
-        Local,     // sent from a local interface, including loopback
-        OriginTag, // payload of CMD_ORIGIN_TAG
+        Broadcast, // received directly from original client as bcast/mcast, process directly
+        Forwarding,// received directly from original client as ucast, forward
+        Forwarded, // forwarded by a local peer
+        OriginTag, // ... with CMD_ORIGIN_TAG
     };
 
-    void process_one(const SockAddr& dest, const uint8_t* buf, size_t nrx, origin_t origin);
-    static void handle_static(evutil_socket_t fd, short ev, void *raw)
-    {
-        (void)fd;
-        auto self = static_cast<UDPCollector*>(raw);
-        try {
-            log_debug_printf(logio, "UDP %p event %x\n", self->rx.get(), ev);
-            if(!(ev&EV_READ))
-                return;
+    void process_one(const uint8_t* buf, size_t nrx, origin_t origin,
+                     const IfaceMap::Current& ifinfo);
+    static void handle_static(evutil_socket_t fd, short ev, void *raw);
 
-            // bound number of packets handled before going back to the reactor
-            for(unsigned i=0; i<16 && self->handle_one(); i++) {}
-
-        }catch(std::exception& e) {
-            log_crit_printf(logio, "Ignoring unhandled exception in UDPManager::handle(): %s\n", e.what());
-        }
-    }
-
-    void forwardM(const SockAddr& origin, const uint8_t* buf, size_t len);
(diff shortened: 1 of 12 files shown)


mdavidsaver:
While I hope this PR will be a practical no-op for the vast majority of hosts, I see a significant chance of some serious regression or interoperability issue.  So this PR will likely "age" a bit until I can do (or receive) some test results with the various other PVA implementations (core.pva, pvAccessCPP/Java).

So far I have done some interop testing with phoebus.

src/udp_collector.cpp
```diff
@@ -261,16 +260,24 @@ bool UDPCollector::handle_one()
         return true;
     }
 
-    log_hex_printf(logio, Level::Debug, rxbuf, nrx, "UDP Rx %d, %s -> %s @%u (%s)\n",
-            nrx, src.tostring().c_str(), dest.tostring().c_str(), unsigned(rx.dstif), bind_addr.tostring().c_str());
+    origin_t origin = Direct;
+    if(manager->ifmap.is_lo(rx.dstif) && dest.compare(lo_mcast_addr.addr,false)==0) {
+        // packet forwarded by a local PVA peer (maybe us) as IPv4 local multicast
+        origin = Forward;
```
mdavidsaver, line 266: @FreddieAkeroyd This is the change to detect a "forwarded" packet, arriving via a loopback interface, with the (multicast) destination address of `224.0.0.128`.  This is distinct from eg. a packet arriving by loopback destined for `127.0.0.1` or `127.255.255.255`.

With this PR, it is possible to achieve the same effect as the Linux specific `127.255.255.255` in a portable way.

```sh
EPICS_PVA_ADDR_LIST=127.0.0.1 \
EPICS_PVA_AUTO_ADDR_LIST=NO \
  ./bin/linux-x86_64/pvxget from:ioc1 from:ioc2
```

However, this will need some interop testing to make sure that other implementations treat this situation sensibly.

FreddieAkeroyd:
Hi @mdavidsaver i tried this with a `127.255.255.255` address list and see messages like `WARN pvxs.udp.io Ignore originated from 127.255.255.255 TT` reported to the client and in all IOC logs, with `127.0.0.1` I get no warning. Prior to the change i had no warnings with 127.255.255.255 address list. In all cases I get the PV read OK.  With this PR change should I have no need to ever use 127.255.255.255 for searches?

mdavidsaver:
> Hi @mdavidsaver i tried this with a `127.255.255.255` address list

Could you try `127.0.0.1` in the address list?

> With this PR change should I have no need to ever use 127.255.255.255 for searches?

I don't think so.

mdavidsaver:
While `pvxlist` works fine, `pvlist` (from pvAccessCPP) does not.

It looks like `pvlist` is incorrectly sending a broadcast search with the Unicast flag set.  sigh...

```
2025-02-19T18:38:35.647184195 WARN pvxs.udp.io Ignore originated from 10.127.127.255 TT
2025-02-19T18:38:35.987015530 WARN pvxs.udp.io Ignore originated from 10.127.127.255 TT
```

pvlist-unicast

FreddieAkeroyd:
With 127.0.0.1 address list:

pvxget works fine

pvget works but i get the following printed in the ioc log each time i run it
ERR pvxs.tcp.io connection to Client 127.0.0.1:55743 closed with socket error 10054 : An existing connection was forcibly closed by the remote host.

pvlist, pvxlist work fine, no warning/error messages; - however just after printing all its output pvxlist pauses for about 5 second before giving the prompt back whereas pvlist gives the prompt back immediately after output 


With 127.255.255.255 address list:

pvget works with same socket close error as above appearing in ioc log

pvxget prints `WARN pvxs.udp.io Ignore originated from 127.255.255.255 TT` at both client window and all running iocs, but reads value ok

pvlists results in a `WARN pvxs.udp.io Ignore originated from 127.255.255.255 TT` in server logs

pvxlist is like pvxget with warning at both client and ioc windows;  the same 5 second pause as for 127.0.0.1 is present

mdavidsaver:
Updated again.  This has turned into a larger change due to a `IfaceMap` design change (the network interface info cache).

mdavidsaver:
@FreddieAkeroyd If you have time, I would appreciate a re-test.

> pvxlist pauses for about 5 second before giving the prompt back whereas pvlist gives the prompt back immediately after output

This is an intentional difference.  `pvxlist` got a longer default timeout when it learned how to listen for beacons.  You could pass eg. `-w 1` if you are feeling impatient :)

The other log messages you mention should not occur normally.  "Ignore originated" might happen briefly if you bring up a new network interface while the process is running.  "closed with socket error 10054" should only happen if the peer exits abnormally.

And, of course, there should be no new misbehavior...

FreddieAkeroyd:
Hi @mdavidsaver will try and get to this next week, i've had rather a lot of admin to do these last few weeks but that is now tailing off

mdavidsaver:
Applied as of 190eb8750f0d31422e621bb95df7b32a0c3ab906

FreddieAkeroyd:
@mdavidsaver thanks for this and for pushing https://github.com/epics-base/pvxs/commit/e30640a6f98af3d681401d23be6348f2abc9ce18

PR #102, opened 2025-02-20
Process requests targetting broadcast address

FreddieAkeroyd:
I was running the following setup:

machine A - EPICS_PVA_ADDR_LIST=subnet_broadcast_address
machine B - p4p gateway on main network, IOCs bound to 127.0.0.1, p4p using 127.255.255.255 as EPICS_PVA_ADDR_LIST

I was unable to resolve IOC PVs on machine B from machine A until I made the enclosed changes

### DIFF
--- a/src/udp_collector.cpp
+++ b/src/udp_collector.cpp
@@ -390,6 +390,14 @@ void UDPCollector::process_one(const SockAddr &dest, const uint8_t *buf, size_t 
             for(auto L : listeners) {
                 if(L->searchCB && (L->dest.addr.isAny() || L->dest.addr==dest)) {
                     (L->searchCB)(*this);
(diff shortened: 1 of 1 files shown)


mdavidsaver:
This should happen earlier during the setup phase.

https://github.com/epics-base/pvxs/blob/f2e49a88445fd6f352cd7ec75d5c0f673756f43e/src/server.cpp#L419

Actually, I'm wondering if my comment here is incorrect wrt. "not winsock"?

https://github.com/epics-base/pvxs/blob/f2e49a88445fd6f352cd7ec75d5c0f673756f43e/src/server.cpp#L444-L455

FreddieAkeroyd:
Hi @mdavidsaver  removing `evsocket::ipstack!=evsocket::Winsock` fixed the issue for me, thanks

PR #108, opened 2025-03-28
Changed Float32/Float64 to Float32A/Float64A in the NTNDArray TypeDef

joaopaulosm:
The union type of the `value` field in the `NTNDArray` definition is set to array types for all but `Float32` and `Float64`. 
Shouldn't they be also array types? 

I'm preparing some unit tests, but `make runtests` passed (for my Ubuntu machine).


mdavidsaver:
Indeed this is a bug...

PR #110, opened 2025-04-04
allow to  have pvxs in arbitrary location

dirk-zimoch:
`bundle/usr` gets installed into `$(INSTALL_LOCATION)` but was searched for relative to the PVXS source directory, which only works under specific circumstances.


anjohnson:
Core Group: Problem is related to when this module needs to build & install its bundled libevent, possibly only when this is added as a git submodule in `$(EPICS_BASE)/modules`.

NB that libevent is a git submodule of this module.

AI on @anjohnson to take a look.

PR #112, opened 2025-04-10
LIBEVENT hint for homebrew

graeme-winter:
For macOS / homebrew install of `libevent` give the reader a hint on how to build, now updated to recognise that the brew installation prefix is configurable.

### DIFF
--- a/documentation/building.rst
+++ b/documentation/building.rst
@@ -35,7 +35,7 @@ On Debian/Ubuntu. ::
 On macOS / Homebrew. ::
 
     brew install libevent
-    export LIBEVENT=/opt/homebrew
+    export LIBEVENT=$(brew --prefix)
 
 To build from source (Requires `CMake <https://cmake.org/>`_): ::
 
--- a/documentation/building.rst
+++ b/documentation/building.rst
@@ -32,6 +32,11 @@ On Debian/Ubuntu. ::
 
     apt-get install libevent-dev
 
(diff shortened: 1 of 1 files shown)


anjohnson:
@dirk-zimoch did you see this PR? Does it impact what you were working on earlier today?

graeme-winter:
> > Does it impact what you were working on earlier today?
> 
> Yes, kind of. My plan is to reference homebrew in base so that all modules have it in their -I and -L paths. Now I know about brew —prefix.

In fairness that would also resolve my issue, beyond the need for `brew install libevent`

Will be happy to test anything when the need arises.

mdavidsaver:
A note on how `$(LIBEVENT)` (the Make variable) is consumed.

https://github.com/epics-base/pvxs/blob/e32a752d12b9cc417b7337ff178dd54d36b4a8f3/configure/CONFIG_PVXS_MODULE#L11-L16

The plain `LIBEVENT` is intended as a shorthand when no cross compilation will happen.

Although both will work, I find it preferable to put values like this into `epics-base/configure/CONFIG_SITE.local` rather than environment variables.  (these days it is possible to build Base and support modules without defining any environment variables)

And a further note for the future.  A future PVXS release (likely 1.4.0) will include a change to capture the LIBEVENT (and eventually OPENSSL) locations during the PVXS build.  Then it will only be necessary to set these variables when building PVXS itself, and not for dependent modules/applications.

PR #120, opened 2025-04-30
fix: Fixing how the hostname is identified to consider IPv6

### LINKED ISSUE #117: pvSecurity does not work when using host (HAG)
**Description**

When using pvxs, pvSecurity seems to not be working with host definition (HAG), for hostname or IP address.
The issue happens only for PVA, for CA it works fine.

**To Reproduce**

1. create the follow files, replacing $(HOSTNAME) by the host name or IP of where IOC is running

```
$ cat sec.acf 
HAG(hag) {$(HOSTNAME)}               
ASG(DEFAULT) {                                    
  RULE(1,READ)                                    
  RULE(1,WRITE) {                                 
    HAG(hag)                                      
 }                                                
}     

$ cat simple.db                       
record(ai, "MUMBLE") {                                                                         
    field(VAL, "12")                                                                           
}

$ cat st.cmd 
var asCheckClientIP 1                                                     
asSetFilename("$(PWD)/sec.acf")                                    
dbLoadRecords("simple.db")                                                
```

2. Run the ioc:
`softIocPVX st.cmd`

3. Try to change the pv value
`pvxput MUMBLE 3`

**Expected behavior**
The value should change

**Information (please complete the following):**
 - PVXS Version or Git commit ID: cd5570b35035994d87c3c6170ebabcda26359ea0
 - EPICS Base Version: 7.0.9
 - libevent Version: 2.0.21
 - EPICS_HOST_ARCH: linux-x86_64
 - Host OS: CentOS 7
 - Compiler version: 4.8.5

Alternately, from a successful build, include the output of `pvxinfo -D`.

**Network configuration**
No additional configuration

gabrielfedel:
This PR fix the host identification, which right now considers only IPv4 addressing.
Thanks to @tboegi who pointed the different addressing ways

Fixes #117


ioc/credentials.cpp
```diff
@@ -24,9 +24,26 @@ namespace ioc {
...
-    auto pos = clientCredentials.peer.find_first_of(':');
-    host = clientCredentials.peer.substr(0, pos);
+    // There are 3 possibilities of addresses
+    // IPv4 wrapped as IPV6 : e.g.: [::ffff:192.168.1.1]:9999
+    // IPv6: e.g. [2001:0db8:85a3:0000:0000:8a2e:0370:7334]:9999
+    // IPv4 only: e.g.: 192.168.1.1:9999
+    auto pos = clientCredentials.peer.find_first_of("[::ffff:");
+    auto end = clientCredentials.peer.find_first_of(']');
+    //IPv4 wrapped as IPv6
+    if (pos != clientCredentials.peer.npos) {
```
tboegi, line 34: Question: What happens, if the address does not include "[::ffff:" ?
Will pos be null ?

gabrielfedel: The result of find_first_of will be npos https://en.cppreference.com/w/cpp/string/basic_string/find_first_of

tboegi: Nice - excuse my missing knowledge in C++
The patch then looks good to me.

gabrielfedel: No problem! Thanks for reviewing!

ioc/credentials.cpp
```diff
@@ -24,9 +24,26 @@ namespace ioc {
...
+    }
+    //IPv6
+    else if (end != clientCredentials.peer.npos)
+        host = clientCredentials.peer.substr(1, end-1);
+    //IPv4
+    else {
+      // Extract host name part (or whole thing if no colon present)
+      pos = clientCredentials.peer.find_first_of(':');
+      host = clientCredentials.peer.substr(0, pos);
+    }
```
mdavidsaver, line 46: Did you find `evutil_parse_sockaddr_port()` and `evutil_inet_ntop()`?

tboegi: No, we didn't. Question: 
@mdavidsaver  How do you think that these functions could help us ?

mdavidsaver: `evutil_parse_sockaddr_port()` will parse `<host>:<port>`, considering the various forms of IPv4/6 addresses.  `evutil_inet_ntop()` renders an IPv4/6 address as a string.

As I think about it though, it would be simpler to use `pvxs::SockAddr`, which already knows about IPv6 address encodings.  The printer for that type will omit the port number if it is zero.

```c++
SockAddr addr(clientCredentials.peer);
addr.setPort(0);
std::string host(SB()<<addr);
```

gabrielfedel: So, the solution using `pvxs::SockAddr`  gets the right address, but it seems it cannot handle IPv6 addresses, or at least cannot handle this kind of address:
`[::ffff:192.168.1.1]` or `::ffff:192.168.1.1`
For the current code work it is needed to unwrap the IPv4 

I will try to understand better what is going on for these IPv6 addressing not working

mdavidsaver: As it happens, I think I have a way to handle this from another branch.  I have pushed a commit which adds `SockAddr::map6to4()` to "unwrap" IPv4 addresses encoded as IPv6.

mdavidsaver:
oops... I also have one usage of `ipAddrToDottedIP()`, which is not aware of IPv6.

https://github.com/epics-base/pvxs/blob/cd5570b35035994d87c3c6170ebabcda26359ea0/src/server.cpp#L695

ioc/credentials.cpp
gabrielfedel: Ah great, I tested and it seems to work fine! I changed my commit ut use the map6to4

mdavidsaver:
Applied as of b0b8d60656e98fe0ec62aa91e71c067efc0dd6e2, with additional fix in 839fc01bfd1e52e6249bf7e3d5cbc6d418379815

PR #121, opened 2025-05-05
Rename generated CONFIG_SITE to TOOLCHAIN

mdavidsaver:
No longer generate `configure/CONFIG_SITE.Common.$(T_A)`, which should never be generated...

Instead, generate and install `$(INSTALL_LOCATION)/cfg/TOOLCHAIN_PVXS.$(T_A)` in addition to the `cfg/CONFIG_PVXS_*` and `cfg/RULES_PVXS_*` files.

These generated `TOOLCHAIN_PVXS.*` files are consumed by `cfg/CONFIG_PVXS_MODULE`.

The contents will look like the following.  So the location of libevent, and its dependencies, are captured at PVXS build time.  

```
$ cat cfg/TOOLCHAIN_PVXS.linux-x86_64
LIBEVENT_PREFIX_linux-x86_64 = 
LIBEVENT_BUNDLE_LIBS_linux-x86_64 = event_openssl event_core event_pthreads
LIBEVENT_SYS_LIBS_linux-x86_64 =

$ cat cfg/TOOLCHAIN_PVXS.linux-x86_64-debug 
LIBEVENT_PREFIX_linux-x86_64-debug = /home/dev/epics/pvxs/bundle/usr/linux-x86_64-debug
LIBEVENT_BUNDLE_LIBS_linux-x86_64-debug = event_core event_pthreads
LIBEVENT_SYS_LIBS_linux-x86_64-debug =
```

Otherwise, this change should be transparent.  Of course, I always suspect that any Makefile change will cause problems for someone.


mdavidsaver:
I have been sitting on this change for some time on the secure PVA development branches.  In part because it is not an essential fix, and I am not certain whether to consider this build system change like an ABI change.  I am thinking to include it now since fixing https://github.com/epics-base/pvxs/pull/120#issuecomment-2852396511 will trigger 1.4.0 .

PR #130, opened 2025-09-11
fix INSTALL_LOCATION!=TOP

mdavidsaver:
Attempts to get `make INSTALL_LOCATION=/some/other/dir` to work correctly.  Including `bundle/`.

Contains the equivalent of #110 after b0b8d60656e98fe0ec62aa91e71c067efc0dd6e2.  It think it also moots #124 by avoiding the `$(error` assertions in TOP, `configure/`, and `setup/` when the contents of `$(INSTALL_LOCATION)/cfg/` are in flux.

@simon-ess @dirk-zimoch @jeonghanlee fyi.


mdavidsaver:
Applied as 38cc6deb74e05b2a75550303064aed9c8876752f

PR #133, opened 2025-09-20
OperationBase::name() segfault when called immediately after Operation creation

m2es3h:
On construction of an OperationBase instance, `std::shared_ptr<Channel> OperationBase::chan` is nullptr until the Channel is constructed. Channel construction happens in a lambda function passed to `context->tcp_loop.dispatch([](){ chan = Channel::Build(...)})`.

In the time between OperationBase construction and deferred Channel construction, calling `OperationBase::name()` dereferences the empty shared_ptr to get `chan->name`.

This fix checks for nullptr first and returns a `std::string&` to a default string instead if needed.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -230,7 +230,7 @@ Connect::~Connect() {}
 
 ConnectImpl::~ConnectImpl() {}
 
-const std::string& ConnectImpl::name() const
+const std::string ConnectImpl::name() const
 {
     return _name;
 }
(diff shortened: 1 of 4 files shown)


mdavidsaver:
Confirmed.

mdavidsaver:
Not crashing is certainly an improvment!

Really though, `Operation::name()` should be able to return the channel name immediately.  Indirectly to the `Channel` could be seen as a premature optimization on my part.  And I was not even consistent (cf. `SubscriptionImpl::channelName`).  So I think the straightforward solution would be to store a channel name directly in OperationBase.

m2es3h:
> ... So I think the straightforward solution would be to store a channel name directly in OperationBase.

Makes sense, I've updated the PR to do this. I've also changed the return types of `name()` to copies instead of references, since using string references that outlive the OperationBase would also cause a segfault.

PR #159, opened 2026-03-25
Reduce transient memory allocations

mdavidsaver:
As an prospective micro-optimization, reduce the number of short lived / transient allocations.  Principally from `std::string` and `BitMask` (containing a `std::vector`).

c++17 std::string_view interface, std::from_chars for parsing, and the associated "transparent comparator" feature for std::map.  Also switches `BitMask` from dynamic to a fixed maximum capacity of 256 bits.  Surely no one would ever need more, right?

Analysis using systemtap https://github.com/mdavidsaver/ltrace/blob/master/alloc_time.stp

Testing a single subscriber to a 1Hz NTScalar double.

Initially, 8x one seciond long lived allocations (updates to subscription cache) vs. 71 shorter lived allocations (8us -> ~512us).

```
Run for 4813878 us
untracked 2
tracked 83 allocs min 8 avg 96622 max 1001133
 value |-------------------------------------------------- count
     2 |                                                    0
     4 |                                                    0
     8 |@@@@@@@@@@@@@@@@@@@@@@@                            23
    16 |@@@@@@@@@@@                                        11
    32 |@@@@@@@@@@@                                        11
    64 |@@@@                                                4
   128 |@@@                                                 3
   256 |@@@@@@@@@@@@@@@@@@@                                19
   512 |@@@@                                                4
  1024 |                                                    0
  2048 |                                                    0
       ~
131072 |                                                    0
262144 |                                                    0
524288 |@@@@@@@@                                            8
1048576 |                                                    0
2097152 |                                                    0
```

With this branch 10x longer vs. 24x shorter.

```
Run for 5484305 us
untracked 3
tracked 40 allocs min 163 avg 250487 max 1001064
  value |-------------------------------------------------- count
     32 |                                                    0
     64 |                                                    0
    128 |@@@@@@                                              6
    256 |@@@@@@@@@@@@@@@@@@@@@@@@                           24
    512 |                                                    0
   1024 |                                                    0
        ~
 131072 |                                                    0
 262144 |                                                    0
 524288 |@@@@@@@@@@                                         10
1048576 |                                                    0
2097152 |                                                    0
```

Now the shortest allocations (~200us) are from libevent buffering and Values from the subscription update queue.  In line with my expectations.

### DIFF
--- a/configure/toolchain.c
+++ b/configure/toolchain.c
@@ -14,7 +14,7 @@ CONFIG_LOADED=YES
 #if __GNUC__
 /* also true for clang */
 GNUISH=YES
-USR_CXXFLAGS += -std=c++11
+USR_CXXFLAGS += -std=c++17
 #endif
 
 #if __GNUC__ && !__clang__
--- a/setup.py
+++ b/setup.py
@@ -604,8 +604,8 @@ def define_DSOS(self):
     cxx11_flags = []
     if probe.try_compile('int probefn() { auto x=1; return x; }',
                          language='c++',
-                         extra_preargs=['-std=c++11']):
-        cxx11_flags += ['-std=c++11']
+                         extra_preargs=['-std=c++17']):
+        cxx11_flags += ['-std=c++17']
 
     pvxs_abi = '%(PVXS_MAJOR_VERSION)s.%(PVXS_MINOR_VERSION)s'%pvxsversion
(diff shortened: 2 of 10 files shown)


mdavidsaver:
> As an prospective micro-optimization ...

It is not clear to me whether avoiding these allocations will be worth the effort of integrating these changes in some form.

anjohnson:
> whether avoiding these allocations will be worth the effort of integrating

Ack, one other quick test you could try would be to see how the max data bandwidth changes.

PR #162, opened 2026-04-02
doc: show how to create arrays of NTScalar

exzombie:
This is not clear from the docs right now. A small change that I hope will make it more obvious.


mdavidsaver:
Thanks.

Issue #161, opened 2026-04-01
Server evbuffer grows without bound for slow monitor clients with large payloads


kangarooMouse:
**Describe the bug**
When a PVXS server posts large values (e.g., NTNDArray image data, ~10 MB per update) via SharedPV::post() to multiple monitor subscribers, the libevent output buffer (evbuffer) for slow clients grows without bound, eventually exhausting server memory.

**To Reproduce**
1. Large area detector at full rate (5Gbps), NTNDArray Shared PV, 10Gbps NIC
2. Connect 1 client, fine
3. Connect 2 client, borderline fine
4. Connect 3 clients, server OOM

**Expected behavior**
Add clients, NIC saturates, client framerates drop, server is stable.

**Information (please complete the following):**
 - PVXS Version or Git commit ID: 6f7942e
 - EPICS Base Version: 7.0.6
 - libevent Version: libevent-2.1.12-8
 - EPICS_HOST_ARCH: linux-x86_64
 - Host OS: Rocky 9
 - Compiler version: 11.5.0

**Additional context**
I rebuilt pvxs with the following changes, and am the first to admit naïveté but the results look good. 

src/servermon.cpp — Two fixes to doReply()

Fix 1 (line 225-234): The self-reschedule at the end of doReply() now checks evbuffer_get_length(tx) <
tcp_tx_limit before dispatching directly. If the TX buffer is full, it routes to conn->backlog instead,
which bevWrite() drains when space opens up.

Fix 2 (lines 132-140): Before serialization, if state is Executing and the TX buffer is already at/above the
 limit, doReply() sets scheduled = true and defers to backlog, returning immediately without serializing
another frame.

I'd like to get your take on these changes.

mdavidsaver:
sigh...  Probably related to https://github.com/epics-base/pvxs/issues/136 (see https://github.com/epics-base/pvxs/commit/4249885f8e8c1b04c9ce1b30b386ea2d099a2be1).

mdavidsaver:
The line numbers quoted seem to be a bit off.  fyi. the github.com web ui has a way of constructing links referencing commit, file, and line(s).  I think this is the first range mentioned.

https ://github.com/epics-base/pvxs/blob/6f7942e4af38a0b84630deb837f2b7bccf53a62a/src/servermon.cpp#L210-L218

https://github.com/epics-base/pvxs/blob/6f7942e4af38a0b84630deb837f2b7bccf53a62a/src/servermon.cpp#L210-L218

[...]

PR #164, opened 2026-04-06
server: monitor TX check buffer level on each iteration

mdavidsaver:
Attempt to address #161.  Move check on connection buffer from `maybeReply()`, run when deciding whether to start draining subscription queue, into the beginning of `doReply()`.  There it will be run on each iteration.

Running only the first time permits a situation where a busy publishing can grow the queue without bound by keeping `doReply()` from idling.


mdavidsaver:
This change still looks ok to me.

PR #167, opened 2026-04-15
Add USE_BUNDLE option

FreddieAkeroyd:
If `BUNDLE_BUILD` is set to `YES` via e.g. `CONFIG_SITE.local` then a `make` at top level will build `libevent` from `bundle` directory without it needing to be built separately first; `bundle` is also added to the clean target and DLLs are installed to pvxs bin directory so the libevent bin area does not need to be separately added to windows `PATH`

### DIFF
--- a/bundle/Makefile
+++ b/bundle/Makefile
@@ -99,6 +99,12 @@ help:
 	$(ECHO) " # remove one target"
 	$(ECHO) " make -C bundle clean.<epics-arch> # remove for one target"
 
+install: libevent
+
+uninstall: clean
+
+runtests:
+
 libevent: libevent.$(EPICS_HOST_ARCH)
 
 clean:
--- a/src/Makefile
+++ b/src/Makefile
@@ -111,6 +111,10 @@ endif
 
 LIB_SYS_LIBS += $(LIBEVENT_SYS_LIBS)
 
+ifeq ($(USE_BUNDLE),YES)
+BIN_INSTALLS_WIN32 += $(wildcard $(LIBEVENT_PREFIX)/lib/*.dll)
+endif
+
 #===========================
 
 include $(TOP)/configure/RULES
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,11 @@ include $(TOP)/configure/CONFIG
 # Directories to build, any order
 DIRS += configure
 
+ifeq ($(USE_BUNDLE),YES)
+DIRS += bundle
+configure_DEPEND_DIRS = bundle
+endif
+
 DIRS += setup
 setup_DEPEND_DIRS = configure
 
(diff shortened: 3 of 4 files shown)


src/Makefile
```diff
@@ -111,6 +111,10 @@ endif
 
 LIB_SYS_LIBS += $(LIBEVENT_SYS_LIBS)
 
+ifeq ($(USE_BUNDLE),YES)
+BIN_INSTALLS_WIN32 += $(wildcard $(LIBEVENT_PREFIX)/lib/*.dll)
+endif
```
mdavidsaver, line 116: While I understand the motivation of avoiding `%PATH%` manipulation, I don't like mixing together libevent and PVXS binaries.

mdavidsaver (commented):
Ok in principle.

mdavidsaver:
Taking a step back, how is `%PATH%` constructed for DLL builds in the ISIS environment(s)?

FreddieAkeroyd:
this is done by usual EPICS `dllPath.bat` which searches all bin directories specified via modules specified in the `RELEASE` files. I could add the LIBEVENT directory to RELEASE, but i copied the DLLs so it only needed to have PVXS specified. Should it find LIBEVENT via one of the `cfg` when generating `dllPath.bat` ?

mdavidsaver:
> this is done by usual EPICS `dllPath.bat` ...

I have been working with EPICS for awhile now.  So I was surprised to find myself asking; what is `dllPath.bat`?  The answer apparently is something I never see in my own projects because I run `makeBaseApp` on Linux, and this TARGET is only included when expanding for a windows target.  I learn something new every day.

```
$ git grep dllPath
...
modules/database/src/template/top/exampleBoot/ioc/Makefile@win32:TARGETS = envPaths dllPath.bat
modules/database/src/template/top/exampleBoot/ioc/Makefile@windows:TARGETS = envPaths dllPath.bat
modules/database/src/template/top/iocBoot/ioc/Makefile@win32:TARGETS = envPaths dllPath.bat
modules/database/src/template/top/iocBoot/ioc/Makefile@windows:TARGETS = envPaths dllPath.bat
...
```

bundle/Makefile
```diff
@@ -99,6 +99,12 @@ help:
 	$(ECHO) " # remove one target"
 	$(ECHO) " make -C bundle clean.<epics-arch> # remove for one target"
 
+install: libevent
+
+uninstall: clean
+
+runtests:
```
anjohnson, line 106: The EPICS build system can recursively request more build targets than just these for any directory entry named in the parent `Makefile`'s `DIRS` variable, depending on what the user asked it to build. The complete set of targets that can get passed to subdirectory `Makefiles` via recursive `$(MAKE) -C` calls is:

1. `inc`
2. `build`
3. `install`
4. `buildInstall`
5. `clean`
6. `realclean`
7. `archclean`
8. `runtests`
9. `tapfiles`
10. `clean-tests`
11. `test-results`
12. `junitfiles`
13. `<arch>` (any architecture name, e.g. `windows-x64-debug`)
14. `<action>.<arch>` (any of the 12 actions above dotted with an arch, e.g. `clean.windows-x64-debug`)

PR #168, opened 2026-04-16
handle INSTALL_LOCATION!=$(TOP)

mdavidsaver:
Due to ordering of make fragment includes in epics-base (https://github.com/epics-base/epics-base/issues/859), attempts to set INSTALL_LOCATION!=$(TOP) from within a fragment will change the ordering of cfg/ includes in a way which causes mismatches between CONFIG_PVXS_MODULE and RULES_PVXS_MODULE.  eg. when setting `INSTALL_LOCATION` from configure/CONFIG_SITE.local

Resolve this by including CONFIG_PVXS_MODULE and TOOLCHAIN_PVXS directly from the source or O.$(T_A) directory.

Should be transparent to dependant modules.

A variation on the second idea mentioned in https://github.com/epics-base/epics-base/issues/859#issuecomment-4255811531 .  Adds special handling of `CONFIG_PVXS_MODULE` only.  This is different than the state prior to b0b8d60656e98fe0ec62aa91e71c067efc0dd6e2, when `RULES_PVXS_MODULE` was also specially included.


mdavidsaver:
superseded by #170.

PR #115, opened 2025-04-20
Secure PVAccess

george-mcintyre:
All changes for Secure PVAccess.  

This PR is not meant to be merged.

PR #171, opened 2026-04-18
Secure PVAccess with certificate status monitoring support

george-mcintyre:
This is the latest version of Secure PVAccess that contains support for status monitoring

edit: MD: previously... #115 #53


george-mcintyre:
Remove all code under examples (docker), and in ci, and yaml in ci.  docker-pub.yml, tag-spva.yml

.github/workflows/release.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     - uses: actions/checkout@v4
       with:
         fetch-depth: "0"
-        submodules: true
+        submodules: recursive
```
mdavidsaver, line 21: Revert?

.github/workflows/release.yml
```diff
@@ -41,6 +41,7 @@ jobs:
 
     - name: abi-diff
       run: python .ci/cue.py exec ./abi-diff.sh
+      continue-on-error: true
```
mdavidsaver, line 44: Probably needs a manual fix on merge.

configure/CONFIG_SITE
```diff
@@ -35,13 +35,16 @@ CHECK_RELEASE = YES
...
+
 # These allow developers to override the CONFIG_SITE variable
 # settings without having to modify the configure/CONFIG_SITE
 # file itself.
 -include $(TOP)/../CONFIG_SITE.local
 -include $(TOP)/configure/CONFIG_SITE.local
 
 # MSVC - skip defining min()/max() macros and enable APIs
-USR_CPPFLAGS_WIN32 += -DNOMINMAX -D_WIN32_WINNT=_WIN32_WINNT_VISTA
+USR_CPPFLAGS_WIN32 += -DNOMINMAX -D_WIN32_WINNT=_WIN32_WINNT_VISTA -DNOCRYPT
```
mdavidsaver, line 48: Why "NOCRYPT"?

documentation/index.rst
```diff
@@ -21,12 +21,13 @@ Dependencies
 
 * `EPICS Base <https://epics-controls.org/resources-and-support/base/>`_ >=3.15.1
 * `libevent <http://libevent.org/>`_ >=2.0.1  (Optionally bundled)
+* `openssl <https://openssl-library.org/>`_ >=3.1.0
```
mdavidsaver, line 24: Mark as "optional".

ioc/pvxs/credentials.h
```diff
@@ -0,0 +1,54 @@
...
+namespace pvxs {
+namespace ioc {
+
+/**
+ * eg.
+ * "username"  implies "ca/" prefix
+ * "krb/principle"
+ * "role/groupname"
+ */
+class PVXS_IOC_API Credentials {
```
mdavidsaver, line 28: Demote credentials.h to internal (or remove?).

ioc/pvxs/iochooks.h
```diff
@@ -145,7 +145,6 @@ void testCleanupPrepare();
  *      }
  *      epicsExitCallAtExits();
  *      pvxs::cleanup_for_valgrind();
- *      return testDone();
```
mdavidsaver, line 148: No need to change this comment :)

ioc/credentials.cpp
```diff
@@ -20,25 +20,19 @@ namespace ioc {
...
+ * @param clientCredentials The client credentials to be used for the credentials object
  */
 
 Credentials::Credentials(const server::ClientCredentials& clientCredentials) {
     SockAddr addr(clientCredentials.peer);
     addr.setPort(0);
     host = std::string(SB()<<addr.map6to4());
-
-    // "ca" style credentials
-    if (clientCredentials.method == "ca") {
```
mdavidsaver, line 32: Where does the "ca" special case move to?  Because I think it must be somewhere.

ioc/group.cpp
```diff
@@ -8,16 +8,14 @@
...
+#include <sstream>
 
 #include <epicsThread.h>
 #include <cantProceed.h>
 
 #include "group.h"
 #include "utilpvt.h"
 
-// define printf macro after __attribute__ in event2.h
-#include <epicsStdio.h>
```
mdavidsaver, line 19: Undo.  Need for capture of stdout/err by IOC shell functions.

ioc/Makefile
```diff
@@ -9,18 +9,25 @@
 TOP=..
 
 include $(TOP)/configure/CONFIG
+# cfg/ sometimes isn't correctly included due to an issue in epics-base
+# so we do here (maybe again) as workaround
+-include $(wildcard $(TOP)/cfg/CONFIG*)
```
mdavidsaver, line 14: Undo.  Redundant to https://github.com/epics-base/pvxs/pull/170

ioc/Makefile
```diff
@@ -9,18 +9,25 @@
...
+ifeq ($(EVENT2_HAS_OPENSSL),YES)
+USR_CPPFLAGS += -DPVXS_ENABLE_OPENSSL
+endif
+
 # access to private headers
 USR_CPPFLAGS += -I$(TOP)/src
 USR_CPPFLAGS += -DPVXS_IOC_API_BUILDING
 
 DBD += pvxsIoc.dbd
-CLEANS += pvxsIoc.dbd
```
mdavidsaver, line 21: Undo pvxsIoc.dbd make rule changes.  Now unnecessary.

ioc/pvalink_lset.cpp
```diff
@@ -647,7 +647,7 @@ long pvaPutValueX(DBLINK *plink, short dbrType,
         if(!self->defer) self->lchan->put();
 
         log_debug_printf(_logger, "%s: %s %s %s\n", __func__, plink->precord->name, self->channelName.c_str(), self->lchan->root.valid() ? "valid": "not valid");
-        
+
```
mdavidsaver, line 650: Make another pass to remove whitespace only changes.

src/certstatus.h
```diff
@@ -0,0 +1,950 @@
...
+#include <pvxs/config.h>
+#include <pvxs/log.h>
+#include <pvxs/nt.h>
+#include <pvxs/sharedArray.h>
+
+#include "certdate.h"
+#include "evhelper.h"
+#include "ownedptr.h"
+
+#define CERT_TIME_FORMAT "%a %b %d %H:%M:%S %Y UTC"
```
mdavidsaver, line 33: Unused, remove definition

src/certstatus.h
```diff
@@ -0,0 +1,950 @@
...
+DEFINE_LOGGER(status, "pvxs.certs.status");
+
+// Define permanently valid status time
+#if defined(__TIME_T_MAX__)
+#define PERMANENTLY_VALID_STATUS __TIME_T_MAX__
+#elif defined(__INT_MAX__)
+#define PERMANENTLY_VALID_STATUS (time_t)(__INT_MAX__)
+#else
+#define PERMANENTLY_VALID_STATUS (time_t)(((unsigned long long)~0) >> 1)
+#endif
```
mdavidsaver, line 48: Switch to `std::numeric_limits<time_t>`.

src/certstatus.h
```diff
@@ -0,0 +1,950 @@
...
+// String initializer list
+#define X_IT(name) #name,
+#define O_IT(name) #name,
+#define CERT_STATES {CERT_STATUS_LIST}
+#define OCSP_CERT_STATES {OCSP_CERT_STATUS_LIST}
+
+// Gets status name based on index (portable: GCC/Clang compound literal -> static array)
+inline const char* CERT_STATE(std::size_t index) {
+    static const char* const _names[] = CERT_STATES;
+    return _names[index];
```
mdavidsaver, line 107: Move definition out-of-line to avoid duplication.  Needs bounds checking.  Consider `std::array`.

src/certstatus.h
```diff
@@ -0,0 +1,950 @@
...
+};
+
+// Forward declarations of certificate status structures
+struct CertificateStatus;
+struct PVACertificateStatus;
+
+/**
+ * @brief PVA Certificate status values enum and string
+ */
+struct PVACertStatus : CertStatus {
```
mdavidsaver, line 207: Sub-classes with no added data members or virtual functions?  Why does this exist?

src/certstatus.h
```diff
@@ -0,0 +1,950 @@
...
+ * and the string representation `s` of the value for logging and comparison
+ *
+ * @note This class is not intended to be instantiated directly.
+ * It is used as a base class for PVACertStatus and OCSPCertStatus
+ */
+struct CertStatus {
+    // enum value of the status
+    uint32_t i{0};
+    // string representation of the status
+    std::string s{};
```
mdavidsaver, line 141: Can this be made into an actual `enum`?

Why store the string representation separately when there is (or should be) an `O(0)` lookup of a `const char*`?

src/certstatus.h
```diff
@@ -0,0 +1,950 @@
...
+    }
+
+   protected:
+    /**
+     * @brief Constructor for CertStatus only to be used by PVACertStatus and OCSPCertStatus
+     *
+     * @param status the enum index of the status
+     * @param status_string the string representation of the status
+     */
+    explicit CertStatus(const uint32_t status, std::string  status_string) : i(status), s(std::move(status_string)) {}
```
mdavidsaver, line 193: Passing `std::string` by value is inefficient.

ioc/singlesource.cpp
```diff
@@ -85,11 +86,6 @@ void subscriptionValueCallback(void* userArg, struct dbChannel* pChannel,
     // ARCHIVE events will get the same data fields as VALUE
     if(change & DBE_ARCHIVE)
         change = (change&~DBE_ARCHIVE)|DBE_VALUE;
-
-    // promote DBE_ALARM only to also fetch value
-    if((change & (DBE_VALUE|DBE_ARCHIVE|DBE_ALARM)) == DBE_ALARM)
-        change |= DBE_VALUE;
```
mdavidsaver, line 91: Revert

python/pvxslibs/version.py
```diff
@@ -6,22 +6,21 @@
 import re
 from collections import namedtuple
 
+try:
```
mdavidsaver, line 9: Revert file.

qsrv/Makefile
```diff
@@ -21,6 +21,10 @@ DBD += softIocPVX.dbd
 softIocPVX_DBD += softIoc.dbd
 softIocPVX_DBD += pvxsIoc.dbd
 
+# Ensure correct build order when generating the expanded DBD.
+# pvxsIoc.dbd is produced by ioc/ and installed into $(INSTALL_LOCATION)/dbd.
+$(COMMON_DIR)/softIocPVX.dbd: $(INSTALL_LOCATION)/dbd/pvxsIoc.dbd
```
mdavidsaver, line 26: Unnecessary, revert.

setup/Makefile
```diff
@@ -12,7 +12,7 @@ LIBEVENT ?= $(LIBEVENT_$(T_A))
 LIBEVENT_$(T_A) ?= $(wildcard $(abspath $(INSTALL_LOCATION)/bundle/usr/$(T_A)))
 
 _LIBEVENT_BUNDLE_LIBS_YES = event_openssl
-ifeq (WIN32,$(OS_CLASS))
+ifeq (msvc,$(CMPLR_CLASS))
```
mdavidsaver, line 15: Details?  Are naming differences really MSVC, or a build option of openssl?

src/pvxs/client.h
```diff
@@ -3,6 +3,8 @@
  * pvxs is distributed subject to a Software License Agreement found
  * in file LICENSE that is included with this distribution.
  */
+
+// Test comment to verify non-Docker changes
```
mdavidsaver, line 7: Revert unnecessary deletions in this file.

src/pvxs/client.h
```diff
@@ -1111,6 +1104,10 @@ struct PVXS_API Config : public impl::ConfigCommon {
     inline Config& overrideShareUDP(bool share) { UDP = share; return *this; }
     inline bool shareUDP() const { return UDP; }
 #endif
+
+    //! Config from definitions
+    //! @since UNRELEASED
+    void fromDefs(Config& self, const std::map<std::string, std::string>& defs, bool useenv);
```
mdavidsaver, line 1110: Please restore as static local function.  No benefit to clutter public API.

src/pvxs/config.h
```diff
@@ -0,0 +1,191 @@
...
+     * @note Cannot be set by an environment variable, but is passed in by commandline tools, or set programmatically
+     */
+    double request_timeout_specified{5.0};
+
+
+    /**
+     * @brief the prefix to append to the URI for CREATE, STATUS, ROOT, etc
+     * default "CERT"
+     */
+    std::string cert_pv_prefix{"CERT"};
```
mdavidsaver, line 123: Remove all references to certificate status PV prefix from client and server configuration.  These names must come from certificate extensions exclusively.

src/pvxs/config.h
```diff
@@ -0,0 +1,191 @@
...
+
+    /**
+     * @brief The password for the keychain file
+     */
+    std::string tls_keychain_pwd;
```
mdavidsaver, line 128: I thought we had agreed that this was an anti-feature?  Remove.

src/pvxs/config.h
```diff
@@ -0,0 +1,191 @@
...
+    /**
+     * @brief True if stapling is disabled irrespective of whether TLS is configured
+     */
+    bool tls_disable_stapling{false};
+
+    /**
+     * @brief The request timeout specified in a user call
+     * @note Cannot be set by an environment variable, but is passed in by commandline tools, or set programmatically
+     */
+    double request_timeout_specified{5.0};
```
mdavidsaver, line 116: Currently unused.  Remove.

src/pvxs/netcommon.h
```diff
@@ -42,9 +43,15 @@ struct PVXS_API PeerCredentials {
...
-    //! For "x509" method, common name of the root CA.
+    //! For "x509" method, common name of the root Certificate Authority.
     //! @since UNRELEASED
     std::string authority;
+    // For "x509"
+    // Peer certificate issuer ID (not passed to EPICS security)
+    std::string issuer_id;
+    // For "x509"
+    // Peer certificate serial number (not passed to EPICS security)
+    std::string serial;
```
mdavidsaver, line 54: I do not want to expose issuer/serial through the public API.  Remove.

src/pvxs/nt.h
```diff
@@ -51,18 +51,13 @@ struct Alarm {
 };
 
 /** A scalar, or array of scalars, and meta-data
- *
- * Passing an array TypeCode (eg. Float64A) builds an NTScalarArray.
```
mdavidsaver, line 55: Revert file.

src/pvxs/server.h
```diff
@@ -6,29 +6,30 @@
 #ifndef PVXS_SERVER_H
 #define PVXS_SERVER_H
 
-#include <osiSock.h>
```
mdavidsaver, line 9: osiSock.h __must be included first__ to avoid some windows header ordering issues.

src/pvxs/server.h
```diff
@@ -107,6 +109,10 @@ class PVXS_API Server
     //! Create a client configuration which can communicate with this Server.
     //! Suitable for use in self-contained unit-tests.
     client::Config clientConfig() const;
+    //! Create a client configuration which can communicate with the given Server.
+    //! Suitable for creating inner client context for certificate status monitoring.
+    //! @since UNRELEASED
+    static client::Config clientConfig(const Config &server_config);
```
mdavidsaver, line 115: Redesign to avoid exposing this redundant static method.

src/pvxs/server.h
```diff
@@ -178,7 +184,7 @@ struct PVXS_API Config : public impl::ConfigCommon {
     bool auto_beacon = true;
 
     //! Server unique ID.  Only meaningful in readback via Server::config()
-    ServerGUID guid{};
+    ServerGUID guid;
```
mdavidsaver, line 187: Revert.

src/pvxs/server.h
```diff
@@ -187,26 +193,23 @@ struct PVXS_API Config : public impl::ConfigCommon {
 
     // compat
     static inline Config from_env() { return Config{}.applyEnv(); }
-
     //! Default configuration using process environment
     static inline Config fromEnv()  { return Config{}.applyEnv(); }
+    //! update using defined EPICS_PVA* environment variables
+    Config& applyEnv();
```
mdavidsaver, line 199: Please revert this reordering, the addition of virtual methods to this class, and the exposure of the previously private `fromDefs()` method.

src/pvxs/server.h
```diff
@@ -225,11 +228,27 @@ struct PVXS_API Config : public impl::ConfigCommon {
...
+    //! From definitions
+    //! @since UNRELEASED
+    void fromDefs(Config& self, const std::map<std::string, std::string>& defs, bool useenv);
+
+    Config()
+        : auto_beacon(true)
+        , guid()
+        , BE(EPICS_BYTE_ORDER==EPICS_ENDIAN_BIG)
+        , UDP(true)
+    {}
```
mdavidsaver, line 251: Prefer previous individual member initializers.

src/pvxs/sharedArray.h
```diff
@@ -143,13 +143,13 @@ struct sa_base {
     {}
 
     // build around existing shared_ptr
-    sa_base(const std::shared_ptr<E>& a, size_t len) noexcept
```
mdavidsaver, line 146: I think all changes to this file are due to mis-merge.  Please revert.

src/pvxs/sharedpv.h
```diff
@@ -8,15 +8,16 @@
 #define PVXS_SHAREDPV_H
 
 #include <functional>
+#include <list>
```
mdavidsaver, line 11: Revert changes to this file.

src/pvxs/source.h
```diff
@@ -274,14 +274,7 @@ struct PVXS_API Source {
...
         //! True if the list may change at some future time.
-        bool dynamic = false;
-        List() = default;
-        List(decltype(names)& names, bool dynamic=false)
-            :names(names), dynamic(dynamic)
-        {}
-        List(decltype(names)&& names, bool dynamic=false)
-            :names(std::move(names)), dynamic(dynamic)
-        {}
+        bool dynamic;
```
mdavidsaver, line 277: Revert changes to this file.

src/pvxs/srvcommon.h
```diff
@@ -31,9 +31,11 @@ namespace server {
  * @since 0.2.0
  * @since UNRELEASED Add PeerCredentials base class
  */
-struct ClientCredentials : public PeerCredentials {
+struct PVXS_API ClientCredentials : public PeerCredentials {
     //! (Copy of) Credentials blob as presented by the client.
     Value raw;
+    // For ABI backwards compatibility
+    std::set<std::string> roles() const ;
```
mdavidsaver, line 38: Maintaining ABI backwards compatibility is not a goal.  That goal is rather to document ABI changes.  Remove.

src/pvxs/unittest.h
```diff
@@ -44,7 +44,7 @@ void testSetup();
 PVXS_API
 void cleanup_for_valgrind();
 
-/** A single test case and context messages.
+/** A single test case (or diagnostic line).
```
mdavidsaver, line 47: Changes to this file look like a mis-merge.  Revert.

src/pvxs/util.h
```diff
@@ -169,7 +169,7 @@ struct PVXS_API Detailed {
 
 /** Describe build and runtime configuration of current system.
  *
- * Print information which may be useful when troubleshooting,
+ * Print information which may be using for when troubleshooting,
```
mdavidsaver, line 172: grammar?

src/pvxs/util.h
```diff
@@ -285,13 +285,23 @@ class MPMCFIFO {
...
      * Blocks while queue is empty.
+     *
+     * Backward-compatibility note:
+     * - Public API and semantics are unchanged. This method still returns a T by value
+     *   and wakes writers/readers exactly as before.
+     * - The implementation move-constructs the return value directly from the front() entry
+     *   instead of default-constructing then move-assigning. This removes an incidental
+     *   requirement that T be default-constructible and move-assignable. Only being
+     *   move-constructible is required, which is a relaxation and thus backward compatible
+     *   for previously working types. Types that worked before will continue to work.
```
mdavidsaver, line 296: A clever change.  This note should be a commit message though.

src/certdate.h
```diff
@@ -0,0 +1,498 @@
+/**
+* Copyright - See the COPYRIGHT that is included with this distribution.
+ * pvxs is distributed subject to a Software License Agreement found
+ * in file LICENSE that is included with this distribution.
+ */
+/**
+ * The certificate time functions
```
mdavidsaver, line 7: I think this whole file needs to go away.  I see the need for a pair for functions coverting between `epicsTime` and `ASN1_TIME`.  The rest looks unnecessary to me.  Especially the string parsing and calendar gymnastics!

src/certstatus.h
```diff
@@ -0,0 +1,950 @@
...
+
+    /**
+     * @brief Get the first 8 hex digits of the hex SKID (subject key identifier)
+     *
+     * Note that the given cert must contain the SKID extension in the first place
+     *
+     * @param cert_ptr the cert pointer from which to get the subject key identifier extension
+     * @return first 8 hex digits of the hex SKID (subject key identifier)
+     */
+    static std::string getSkId(const X509* cert_ptr) {
```
mdavidsaver, line 169: Remove?  I think only used for the to be removed `PeerCredential::issuer_id`.

src/certstatus.cpp
```diff
@@ -0,0 +1,723 @@
...
+ * @param asn1_number
+ * @return
+ */
+uint64_t ASN1ToUint64(const ASN1_INTEGER* asn1_number) {
+    if (!asn1_number)                           throw OCSPParseException("ASN1 integer is null");
+
+    const ossl_ptr<BIGNUM> bn(ASN1_INTEGER_to_BN(asn1_number, nullptr), false);
+    if (!bn)                                    throw OCSPParseException("Failed to convert ASN1 integer to BIGNUM");
+    if (BN_is_negative(bn.get()))            throw OCSPParseException("ASN1 integer is negative");
+    if (BN_num_bits(bn.get()) > 64)           throw OCSPParseException("ASN1 integer overflow: value exceeds uint64_t");
```
mdavidsaver, line 276: I do not think it is valid to assume that all serial numbers fall into this range.  Since `serial` will not appear in the public API, I think it will be necessary to store internally as a `BIGNUM`.

src/certstatus.cpp
```diff
@@ -0,0 +1,723 @@
...
+
+    // Extract the basic OCSP response
+    const ossl_ptr<OCSP_BASICRESP> basic_response(OCSP_response_get1_basic(ocsp_response.get()), false);
+    if (!basic_response)                                    throw OCSPParseException("Failed to get basic OCSP response");
+
+    // Verify OCSP response is signed by provided trusted root certificate authority
+    verifyOCSPResponse(basic_response, trusted_store_ptr);
+
+    OCSP_SINGLERESP *single_response = OCSP_resp_get0(basic_response.get(), 0);
+    if (!single_response)                                   throw OCSPParseException("No entries found in OCSP response");
```
mdavidsaver, line 349: Need to handle OCSP with multiple SINGLERESP.  I think `OCSP_resp_find_status()`  does this?

src/certstatus.cpp
```diff
@@ -0,0 +1,723 @@
...
+
+    ASN1_GENERALIZEDTIME *this_update = nullptr, *next_update = nullptr, *revocation_time = nullptr;
+    int reason = 0;
+
+    const OCSP_CERTID *ocsp_cert_id = OCSP_SINGLERESP_get0_id(single_response);
+    const auto observed_cert_id = certIdFromOCSPCertId(ocsp_cert_id);
+    if (observed_cert_id != cert_id)                        throw OCSPParseException(SB() << "OCSP response cert_id mismatch. Expected: " << cert_id << ", Got: " << observed_cert_id);
```
mdavidsaver, line 356: Keeping openssl derived identifiers in their native formats would simplify these sorts of comparisons, and easy my concerns about the fidelity of the conversion processes (eg. truncation of hashs).

src/certstatus.cpp
```diff
@@ -0,0 +1,723 @@
...
+    constexpr int allowed_skew = 300;                      // Allow a 5-minute clock skew
+    const int fallback_max_age = next_update ? -1 : 1800;  // Only enforce a status validity age cap of 30 minutes if `next_update` is missing
+
+    // Check status validity: tolerate skew of 5 minutes and a maximum age provided in next_update or fall back to a max of 30 minutes
+    if (OCSP_check_validity(this_update, next_update, allowed_skew, fallback_max_age) != 1) {
+        const unsigned long err = ERR_get_error();
+        if(err) {
+            char err_buf[256];
+            ERR_error_string_n(err, err_buf, sizeof(err_buf));
+            throw OCSPParseException(SB() << "OCSP_check_validity failed: " << err_buf);
```
mdavidsaver, line 372: Why not `SSLError`?

Should `OCSPParseException` become a sub-class of `SSLError`?

src/certstatus.cpp
```diff
@@ -0,0 +1,723 @@
...
+                                        auto status_update{PVACertificateStatus(update, trusted_store_ptr, cert_id)};
+                                        log_debug_printf(status, "Status subscription %s received: %s\n", s.name().c_str(), status_update.status.s.c_str());
+                                        csm->status_ = std::make_shared<CertificateStatus>(status_update);
+                                        log_debug_printf(status, "Calling (*csm->callback_ref)(status_update)%s\n", "");
+                                        (*csm->callback_ref)(status_update);
+                                        log_debug_printf(status, "Called (*csm->callback_ref)(status_update)%s\n", "");
+                                   } catch (OCSPParseException &e) {
+                                       log_debug_printf(status, "Ignoring invalid %s status update: %s\n", s.name().c_str(), e.what());
+                                   } catch (std::invalid_argument &e) {
+                                       log_debug_printf(status, "Ignoring invalid %s status update: %s\n", s.name().c_str(), e.what());
```
mdavidsaver, line 440: Are there legitimate reasons for these conditions to occur?  Otherwise, these log messages should be promoted to `err`.

src/certstatus.cpp
```diff
@@ -0,0 +1,723 @@
...
+                                   } catch (std::exception &e) {
+                                       log_err_printf(status, "%s\n", e.what());
+                                   }
+                               }
+                           } catch (client::Connected &conn) {
+                               log_debug_printf(status, "Connected Subscription %s: %s\n", s.name().c_str(), conn.peerName.c_str());
+                           } catch (client::Disconnect &conn) {
+                               log_debug_printf(status, "Disconnected Subscription %s: %s\n", s.name().c_str(), conn.what());
```
mdavidsaver, line 448: It might help with understandably if these logs were promoted to `info`.

src/certstatus.cpp
```diff
@@ -0,0 +1,723 @@
...
+}
+
+/**
+ * @brief Get the extension from the certificate.
+ * This method retrieves the extension from the given certificate using the NID_PvaCertConfigURI.
+ * If the extension is not found, it throws a CertConfigNoExtensionException.
+ * @param certificate the certificate to retrieve the extension from
+ * @return the X509_EXTENSION object, if found, otherwise throws an exception
+ */
+X509_EXTENSION *CertStatusManager::getConfigExtension(const X509 *certificate) {
```
mdavidsaver, line 546: All references to this second "config" x509 extension must be removed.

src/certstatus.cpp
```diff
@@ -0,0 +1,723 @@
...
+    }
+
+    // Convert ASN1_INTEGER to BIGNUM
+    const ossl_ptr<BIGNUM> bn(ASN1_INTEGER_to_BN(serial, nullptr), false);
+    if (!bn) {
+        throw CertStatusNoExtensionException("Failed to convert Serial Number to BIGNUM.");
+    }
+
+    // Convert BIGNUM to decimal string
+    char* decimal_str = BN_bn2dec(bn.get());
```
mdavidsaver, line 587: The department of call redundancy is calling.

src/certstatus.h
```diff
@@ -0,0 +1,950 @@
...
+ * @brief Structure representing PVA-OCSP certificate status.  This is a superclass of OCSPStatus
+ *
+ * It contains the OCSP response bytes as well as the date the status was set and how
+ * long the status is valid for.  If the status is revoked, then there is also a
+ * revocation date.  The status field contains the PVA certificate status in numerical and text form.
+ * The ocsp_status field contains the OCSP status in numerical and text form.
+ */
+struct PVACertificateStatus final : OCSPStatus {
+    PVACertStatus status{UNKNOWN};
+    CertDate renew_by{};
```
mdavidsaver, line 436: I do not think that `renew_by` is used anywhere in PVXS.  Remove.

mdavidsaver (changes requested):
I am stopping after accumulating 50 notes in the first ~25% of the code.

mdavidsaver:
I am switching this PR to Draft status pending resolution of https://github.com/slac-epics/pvxs-tls/issues/19 .

configure/CONFIG_SITE
george-mcintyre: NOCRYPT suppresses the legacy wincrypt.h symbols pulled in by windows.h, which otherwise collide with OpenSSL's X509/PKCS7 declarations on MSVC.  Same rationale as the adjacent NOMINMAX.  Added when SPVA introduced the OpenSSL dependency.

.github/workflows/release.yml
george-mcintyre: Agreed. Must be reverted after upstream merge.

setup/Makefile
george-mcintyre: It is MSVC vs unix link-name convention, not an openssl build option. MSVC links libssl/libcrypto by full name; unix uses ssl/crypto (-lssl/-lcrypto). Gated on CMPLR_CLASS and only applied when EVENT2_HAS_OPENSSL.

ioc/pvxs/iochooks.h
george-mcintyre: agreed: fixed in https://github.com/slac-epics/pvxs-tls/commit/4995405

src/pvxs/nt.h
george-mcintyre: agreed: fixed in https://github.com/slac-epics/pvxs-tls/commit/1c71fae

src/pvxs/unittest.h
george-mcintyre: agreed: fixed in https://github.com/slac-epics/pvxs-tls/commit/51a3932

src/pvxs/util.h
george-mcintyre: agreed: fixed in https://github.com/slac-epics/pvxs-tls/commit/53a2c12

documentation/index.rst
george-mcintyre: agreed: fixed in https://github.com/slac-epics/pvxs-tls/commit/6e1264b

ioc/pvalink_lset.cpp
george-mcintyre: agreed: fixed in https://github.com/slac-epics/pvxs-tls/commit/cd4bef0

src/pvxs/config.h
george-mcintyre: agreed: cert-status PV names now come from the certificate extension only. Removed cert_pv_prefix, its accessors, and the EPICS_PVA[S]_CERT_PV_PREFIX handling from client/server config in slac-epics/pvxs-tls@6f24f97. PVACMS, which originates these PVs, now owns the prefix (slac-epics/pvxs-cms@591d753); docs updated in slac-epics/pvxs-docs@d054b2a.

src/pvxs/config.h
george-mcintyre: agreed: the keychain password-file mechanism has been removed (re: https://github.com/epics-base/pvxs/pull/171#discussion_r3334815742). The password is now supplied only via the ;password postfix on the keychain path (EPICS_PVA[S]_TLS_KEYCHAIN=/path;password), or programmatically via setKeychainPassword(). The tls_keychain_pwd member is retained as the store for that postfix password.

Fixed across the stack:

pvxs: slac-epics/pvxs-tls@462e6716
pvxs-cms (same removal for the PVACMS/CA/admin keychains + LDAP, postfix added where missing): slac-epics/pvxs-cms@1c88a2f
pvxs-docs (dev + release): slac-epics/pvxs-docs@6f3c992
phoebus tracked for the matching Java change: ControlSystemStudio/phoebus#3879

src/pvxs/config.h
george-mcintyre: agreed: removed. request_timeout_specified and its get/setRequestTimeout accessors were never read in pvxs; the CLI tools already keep a local timeout for their own wait() calls.

pvxs: slac-epics/pvxs-tls@2622a50c (also removed the dead setRequestTimeout calls in the tools, and monitor's -w, which only fed the dead setter)
pvxs-cms (re-homed the timeout onto its own ConfigCms/ConfigAuthN since it actually uses it): slac-epics/pvxs-cms@24f70fb
pvxs-docs: slac-epics/pvxs-docs@17acf06

src/certstatus.cpp
george-mcintyre: The serial is a 64-bit value by design here, and that assumption is safe because the numeric-serial path is only ever reached for certificates that PVACMS (or another SPVA CMS) issues.

Two things gate it:

We can only monitor certs that tell us where to look. Status monitoring requires the NID_SPvaCertStatusURI extension, which only PVACMS-issued certs carry. getStatusPvFromCert throws CertStatusNoExtensionException for any cert without it (src/certstatus.cpp), so setPeerStatus never establishes a monitor for a foreign cert.
We can only answer status for certs in our DB — i.e. ones we minted. PVACMS generates those serials as uint64 via ASN1_INTEGER_set_uint64 (pvxs-cms/src/common/certfactory.cpp:272) and stores them as 64-bit (serial_number_t = typedef uint64_t, src/openssl.h:41).
Externally-minted certificates are fully supported for TLS trust/identity; that path renders the serial through the BIGNUM-safe string representation (getSerialFromCert → BN_bn2dec, used to build the cert ID at src/certstatus.cpp:606), so a serial of any width works there. Such certs simply don't enter the numeric status-monitoring path, because we neither have their status extension nor can provide status for them.

ASN1ToUint64 (line 276) only parses OCSP responses whose cert-ID string first matches the cert we requested (src/certstatus.cpp:356) — always one of ours — so its 64-bit validation is asserting an invariant that holds by construction, not silently truncating third-party serials.

If we ever wanted to monitor status for certs minted by a different SPVA-aware CA with >64-bit serials, that would be a broader change (widening serial_number_t through the peer-status map, the PVACMS DB column, and the wire schema), which we can take up separately.

ioc/credentials.cpp
george-mcintyre: It moved to src/serverconn.cpp:301-310, the selected=="ca" branch. When the
"ca" auth method is selected, the account is taken from auth["user"] with
everything up to and including the last '/' stripped - the same strip the
constructor used to do. The constructor now just copies the resolved account and
keeps method as its own field, so the "ca" handling lives at the point where the
auth exchange is processed rather than in Credentials.

ioc/singlesource.cpp
george-mcintyre: agreed: reverted in slac-epics/pvxs-tls@c530fda1

src/certstatus.h
george-mcintyre: agreed: the enums are already real enums, and CertStatus::s now stores the
const char* from the existing CERT_STATE/OCSP_CERT_STATE static-table lookup
instead of copying it into a std::string.
- pvxs: slac-epics/pvxs-tls@81e69164
- pvxs-cms: slac-epics/pvxs-cms@14ff941

src/certstatus.h
george-mcintyre: They carry no data or virtuals, but they exist for compile-time type safety
rather than state. PVACertStatus wraps certstatus_t and OCSPCertStatus wraps
ocspcertstatus_t, which gives us three things the bare base couldn't:

- typed comparisons: status == VALID and ocsp_status == OCSP_CERTSTATUS_GOOD
  are checked against the right enum, so a PVA status can't accidentally be
  compared to an OCSP enum (or vice versa).
- distinct member types on CertificateStatus (status vs ocsp_status), so the
  two domains can't be swapped at a call site.
- each constructor selects the correct name table (CERT_STATE vs
  OCSP_CERT_STATE) automatically.

Collapsing them into CertStatus would require the base to carry both enum
comparison overloads, which reintroduces exactly the cross-domain comparison
we're preventing. Given that, I'd prefer to keep them. Happy to revisit if
you'd rather encode the distinction a different way (e.g. a single templated
CertStatus<EnumT>).

python/pvxslibs/version.py
george-mcintyre: agreed: reverted in slac-epics/pvxs-tls@e2a9d7d9

src/pvxs/sharedArray.h
george-mcintyre: agreed: reverted in slac-epics/pvxs-tls@1dbea67b

ioc/group.cpp
george-mcintyre: agreed: restored <epicsStdio.h> (and dropped the unused <sstream>) in
slac-epics/pvxs-tls@24f79c78

qsrv/Makefile
george-mcintyre: agreed: reverted in slac-epics/pvxs-tls@24f79c78

src/certstatus.h
george-mcintyre: agreed: removed the duplicate; CERT_TIME_FORMAT is defined and used in
certdate.h. slac-epics/pvxs-tls@5e835ad2

src/certstatus.h
george-mcintyre: agreed: now (std::numeric_limits<time_t>::max)(). One wrinkle: with a 64-bit
time_t that sentinel is outside gmtime()'s range, which crashed the date
formatter, so CertDate::toString() now treats a null std::tm as a
never-expiring marker instead of dereferencing it.
- pvxs: slac-epics/pvxs-tls@a08fc850
- pvxs-cms: slac-epics/pvxs-cms@3a8601f

src/certstatus.h
george-mcintyre: agreed: the CertStatus constructor now takes const char* instead of a
std::string by value (done as part of the comment-16 change).
slac-epics/pvxs-tls@81e69164

src/certstatus.h
george-mcintyre: agreed: CERT_STATE/OCSP_CERT_STATE are now declared in the header and defined
once in certstatus.cpp using std::array, with a bounds check that returns a
safe fallback for out-of-range indices instead of the name tables being
duplicated into every translation unit that includes the header.
- pvxs: slac-epics/pvxs-tls@dd873c73
- pvxs-cms: slac-epics/pvxs-cms@78f2c50

src/certstatus.h
george-mcintyre: agreed: removed. slac-epics/pvxs-tls@d2f8226c

src/pvxs/netcommon.h
george-mcintyre: agreed: removed from PeerCredentials. pvxs-cms recomputes what it needs from
the kept fields — self-revoke matches the caller's cert CN against the target
serial's recorded CN, and cluster peer checks rely on the CA trust anchor
enforced at the TLS handshake.
- pvxs: slac-epics/pvxs-tls@02e7e9a1
- pvxs-cms: slac-epics/pvxs-cms@17fa607

src/certstatus.h
george-mcintyre: agreed: removed. The only remaining use was the test certificate generator,
which now has its own local helper for reading the subject key identifier.
slac-epics/pvxs-tls@2b34344d

src/pvxs/source.h
george-mcintyre: agreed: reverted in slac-epics/pvxs-tls@47c1eaab

src/pvxs/sharedpv.h
george-mcintyre: agreed: reverted in slac-epics/pvxs-tls@47c1eaab (the header and source are
coupled by the pv_list_t typedef, so both were reverted together)

src/pvxs/server.h
george-mcintyre: agreed: osiSock.h is included first again. slac-epics/pvxs-tls@64ed7391

src/pvxs/server.h
george-mcintyre: agreed: the static clientConfig(const Config&) had no external callers; it is
now a file-local helper in server.cpp and off the public API.
slac-epics/pvxs-tls@64ed7391

src/pvxs/server.h
george-mcintyre: agreed: reverted. slac-epics/pvxs-tls@64ed7391

src/pvxs/server.h
george-mcintyre: agreed: reverted the reordering, updateDefs is non-virtual again, and the
server-side fromDefs() is a file-local free function again.
- pvxs: slac-epics/pvxs-tls@64ed7391
- pvxs-cms (adapted its ConfigCms to the non-virtual base): slac-epics/pvxs-cms@f1deccf

src/pvxs/server.h
george-mcintyre: agreed: restored the individual member initializers. slac-epics/pvxs-tls@64ed7391

src/certstatus.cpp
george-mcintyre: agreed: removed the duplicated conversions. The serial ASN1->decimal and the
issuer key-hash first-8-hex formatting were each repeated (in
certIdFromOCSPCertId, getSerialFromCert and getIssuerIdFromCert); both are now
shared helpers.
- slac-epics/pvxs-tls@31ade1d3
- slac-epics/pvxs-tls@cdc60d8c

(The manual issuer+serial derivation itself is kept rather than read from the
status-PV extension: it has to match the cert ID derived from the OCSP
response's OCSP_CERTID for stapled-response validation.)

src/certstatus.cpp
george-mcintyre: agreed: a status update that fails to parse is not an expected condition, so
these are now logged at err. slac-epics/pvxs-tls@c23262bd

src/certstatus.cpp
george-mcintyre: agreed: promoted to info. slac-epics/pvxs-tls@c23262bd

src/certstatus.cpp
george-mcintyre: OCSPParseException stays a std::runtime_error rather than an SSLError subclass
for two reasons: 
- SSLError lives in openssl.h which already includes
certstatus.h, so deriving from it there would be a circular include; and
- SSLError's constructor drains the OpenSSL error queue (ERR_get_error_all) into
the message, which is wrong for these logical parse/validation errors that are
thrown with no OpenSSL error pending (e.g. "cert_id mismatch"). It would either
add empty noise or splice in unrelated stale errors.

src/certstatus.cpp
george-mcintyre: agreed: removed the second "config" x509 extension and the certificate
configuration PV it fed, end to end. Certificate renewal is unaffected (it
runs off the status PV's renewal_due flag).
- pvxs: slac-epics/pvxs-tls@0b35e904
- pvxs-cms: slac-epics/pvxs-cms@e9e7546
- pvxs-docs: slac-epics/pvxs-docs@5efe517

ioc/pvxs/credentials.h
george-mcintyre: agreed: demoted credentials.h to an ioc-internal header (no longer
installed to the public include tree), and moved PVACMS's access-security
check onto epics-base asLib directly.

- pvxs: reverted ioc/Makefile so pvxs/credentials.h is ioc-internal, and
  removed the stale public copy:
  https://github.com/slac-epics/pvxs-tls/commit/e5f0193b
- pvxs-cms: added src/common/asauth.h (clientCanPut, using asAddClientIdentity
  / asCheckPut directly, with an EPICS_ASLIB_HAS_IDENTITY fallback), replacing
  the ioc Credentials / SecurityClient use in pvacms.cpp and clusterctrl.cpp:
  https://github.com/slac-epics/pvxs-cms/commit/52cc364
- pvxs-docs: rewrote the standalone-server authorization example to call
  asLib directly, and dropped the ioc::Credentials API-reference page and its
  references:
  https://github.com/slac-epics/pvxs-docs/commit/0fb2950

.github/workflows/release.yml
george-mcintyre: agreed: - slac-epics/pvxs-tls@74736928

src/certstatus.cpp
george-mcintyre: agreed: the parser now iterates every SINGLERESP instead of assuming index 0.

- pvxs (`CertStatusManager::parse`) loops over `OCSP_resp_count()` / `OCSP_resp_get0(basic, i)` and
  selects the SINGLERESP whose certificate id matches the expected `cert_id`.
- pvxs-cms (`CmsStatusManager::parse`) diverges: it has no expected `cert_id` in scope, so "find the matching entry" does not apply.  Fails if unexpected number of responses are found.
- https://github.com/slac-epics/pvxs-tls/commit/9ab68165
- https://github.com/slac-epics/pvxs-cms/commit/720030c

src/certstatus.cpp
george-mcintyre: Native comparison is only reachable on the stapling path, where we hold the peer cert and can build an OCSP_CERTID (OCSP_cert_to_id / OCSP_id_cmp). On the status-PV path the identity arrives as a string in the PV name (CERT:STATUS:<issuer_id>:<serial>), with no cert to derive a native CERTID from, so a full switch to native formats is a protocol change rather than a local fix.

src/certdate.h
george-mcintyre: agreed in part: the time_t<->ASN1_TIME conversion pair (asn1TimeToTimeT / timeTToAsn1Time) is now split out into a small certtime.h, and certdate.h delegates to it. The duration parsing (parseDuration / formatDurationMins) is kept because it backs the *_CERT_VALIDITY_MINS config options and the --cert-validity CLI arg, and is covered by testtlstime.t; removing it would drop user-facing behaviour.
- https://github.com/slac-epics/pvxs-tls/commit/3daf3baf
- https://github.com/slac-epics/pvxs-cms/commit/5f67b6d

ioc/Makefile
george-mcintyre: agreed: reverted.  https://github.com/slac-epics/pvxs-tls/commit/e5f0193b

ioc/Makefile
george-mcintyre: agreed: reverted the pvxsIoc.dbd rule changes.  https://github.com/slac-epics/pvxs-tls/commit/e5f0193b

src/pvxs/client.h
george-mcintyre: agreed: reverted the deletions in this file.  https://github.com/slac-epics/pvxs-tls/commit/454e91dc

src/pvxs/client.h
george-mcintyre: agreed: restored as a file-local function.  https://github.com/slac-epics/pvxs-tls/commit/454e91dc

src/pvxs/util.h
george-mcintyre: moved the note into the commit message.  https://github.com/slac-epics/pvxs-tls/commit/23bc9a26

src/pvxs/srvcommon.h
george-mcintyre: agreed: removed the comment.  https://github.com/slac-epics/pvxs-tls/commit/23094f9d

configure/CONFIG_SITE
mdavidsaver: > NOCRYPT suppresses the legacy wincrypt.h symbols pulled in by windows.h, which otherwise collide with OpenSSL

Ok.  Move the `-DNOCRYPT` to a new line and add a comment to this effect.

```make
# supress MS crypto API which conflicts with openssl
USR_CPPFLAGS_WIN32 += -DNOCRYPT
```

src/serverconn.cpp
```diff
@@ -81,16 +109,28 @@ ServerConn::ServerConn(ServIface* iface, evutil_socket_t sock, struct sockaddr *
...
             throw ossl::SSLError("SSL_new()");
+
+        if (!iface->server->tls_context->stapling_disabled && !iface->server->tls_context->status_check_disabled) {
+            try {
+                log_debug_printf(stapling, "Server OCSP Stapling: installing callback%s\n", "");
+                ossl::configureServerOCSPCallback(iface->server, ssl);  // Staple response
+            } catch (certs::OCSPParseException& e) {
+                log_debug_printf(stapling, "Server OCSP Stapling: failed to install callback: %s\n", e.what());
+            } catch (std::exception& e) {
+                log_debug_printf(stapling, "Server OCSP Stapling: failed to install callback: %s\n", e.what());
```
mdavidsaver, line 124: Firstly, different logging points should always print different messages.  Secondly, the `OCSPParseException` case seems entirely redundant to the `std::exception` case anyway.

More significantly, should these failures really log at "debug"?  The message(s) read like "should never happen" situations, which should be logged at "err" or higher so that it is obvious if they somehow do occur.

src/serverconn.cpp
```diff
@@ -81,16 +109,28 @@ ServerConn::ServerConn(ServIface* iface, evutil_socket_t sock, struct sockaddr *
 
 #ifdef PVXS_ENABLE_OPENSSL
     if(iface->isTLS) {
-        assert(iface->server->tls_context);
-        auto ctx(SSL_new(iface->server->tls_context.ctx));
-        if(!ctx)
+        assert(iface->server->tls_context->ctx);
+        auto ssl(SSL_new(iface->server->tls_context->ctx.get()));
```
mdavidsaver, line 113: Also.  Please use RAII to manage this C pointer.

I have sometimes compromised on RAII.  In situations where exception safety can be easily audited.  eg. the original code where `SSL_new()` is followed a single NULL check, then `bufferevent_openssl_filter_new()`.  This change makes the code between complicated enough that it is no longer obvious at a glance to me that no exceptions can propagate from between the two C API calls.

Seeing references to `OCSPParseException` me to suspect that `configureServerOCSPCallback()` does have error cases which should propagate.

src/serverconn.cpp
```diff
@@ -228,81 +269,147 @@ void auth_complete(ServerConn *self, const Status& sts)
...
 
             if(selected=="ca") {
                 auth["user"].as<std::string>([&C, &selected](const std::string& user) {
+            const auto pos = user.find_last_of('/');
+            if ( pos  == std::string::npos) {
+                C->account = user;
+            } else {
+                C->account = user.substr(pos + 1);
+            }
                     C->method = selected;
```
mdavidsaver, line 309: Indentation?

ioc/credentials.cpp
mdavidsaver: > It moved to src/serverconn.cpp:301-310, the selected=="ca" branch ...

Ok.

src/certstatus.h
mdavidsaver: > One wrinkle: with a 64-bit
> time_t that sentinel is outside gmtime()'s range, which crashed the date
> formatter

Maybe an indication to use of a different magic value?  Or a separate "bypass" boolean flag?

src/certstatus.h
mdavidsaver: > Happy to revisit if
> you'd rather encode the distinction a different way (e.g. a single templated
> CertStatus).

I would like to revisit.

src/pvxs/netcommon.h
mdavidsaver: > ... pvxs-cms recomputes what it needs from the kept fields ...

I want to follow up on this statement.

src/certdate.h
mdavidsaver: I need to follow up.

src/certstatus.cpp
mdavidsaver: > ... only ever reached for certificates that PVACMS (or another SPVA CMS) issues.

> We can only answer status for certs in our DB ...

I worry that an assumption is being made that the current pvacms is not only source of certificates with the `SPvaCertStatusURI` extension.

george-mcintyre:
Superseded by #199: same change on a dedicated head branch (slac-epics:community-RC) so review fixes land directly on the branch under review. Review history stays here; all fixes discussed here are included in #199.

PR #172, opened 2026-04-20
client: refresh hostname search destinations

derekste:
## Summary

This updates PVXS client hostname handling so `addressList` and `nameServers` entries remain stateful for the lifetime of a client context instead of being resolved once during config expansion or context creation.

The change preserves original hostname tokens through client config handling, refreshes due hostname entries lazily using DNS TTL/retry policy, and reconnects name-server connections when a hostname resolves to a new address.

## Why

Before this change:

- client config parsing/expansion resolved hostnames and reprinted them as numeric addresses
- `ContextImpl` then built fixed `searchDest` and `nameServers` vectors from those resolved addresses
- long-lived PVXS clients would therefore keep sending searches and using name-server connections based on startup-time DNS results

That meant DNS changes were not reflected until the PVXS client context was recreated.

This PR makes those destinations stateful over the life of the client context, using the existing search and name-server timers rather than adding a background DNS thread.

## Details

- Preserve original hostname tokens through client config handling so they are still available after `Config::expand()`.
- Keep numeric IP, multicast, interface, and auto-discovered address behavior static.
- Add private stateful endpoint handling for client `addressList` and `nameServers`.
- Track the source token, resolved endpoint, resolved/unresolved state, and next expiry/retry time for hostname-backed entries.
- Refresh due UDP search destinations at the start of the existing search path before sending.
- Refresh due TCP name-server destinations before connect/search use.
- If a name-server hostname resolves to a different address, clean up the old connection and let the existing reconnect path build a new one.
- Keep unresolved hostname entries present but fail closed until a later retry succeeds.
- Use DNS TTLs where available on supported POSIX resolver platforms:
  - A-record TTLs for selected IPv4 results
  - AAAA-record TTLs for selected IPv6 results
- Fall back to `300s` after successful lookups without a usable TTL and `60s` before retrying unresolved names.
- Update config tests to reflect that hostnames are preserved rather than always normalized to numeric addresses.

## Relationship To EPICS Base PR #862

This PR is **complementary to**, but does **not require**, [epics-base/epics-base#862](https://github.com/epics-base/epics-base/pull/862).

- This PVXS PR handles **client search/name-server destination statefulness** inside PVXS.
- EPICS Base PR #862 handles **Access Security HAG DNS statefulness and propagation to existing AS clients**.

So there is no new public EPICS Base API dependency here, but the two PRs fit the same larger goal: keeping long-lived PVXS behavior current when DNS changes at runtime.

## Testing

- `make -C pvxs`
- `pvxs/test/O.darwin-aarch64/testconfig`
  - passed `35/35`
- `pvxs/test/O.darwin-aarch64/testnamesrv`
  - passed `5/5`
- `git -C pvxs diff --check`

## Notes

This PR is intentionally scoped to PVXS client `addressList` and `nameServers`. It does not change generic resolver APIs or add a background refresh thread.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -110,10 +110,10 @@ LIB_LIBS += $(LIBEVENT_BUNDLE_LIBS)
 endif
 
 LIB_SYS_LIBS += $(LIBEVENT_SYS_LIBS)
-LIB_SYS_LIBS_Darwin += resolv
-LIB_SYS_LIBS_Linux += resolv
-LIB_SYS_LIBS_freebsd += resolv
-LIB_SYS_LIBS_solaris += resolv
+pvxs_SYS_LIBS_Darwin += resolv
+pvxs_SYS_LIBS_Linux += resolv
+pvxs_SYS_LIBS_freebsd += resolv
+pvxs_SYS_LIBS_solaris += resolv
 
 #===========================
 
--- a/src/Makefile
+++ b/src/Makefile
@@ -110,6 +110,10 @@ LIB_LIBS += $(LIBEVENT_BUNDLE_LIBS)
 endif
 
 LIB_SYS_LIBS += $(LIBEVENT_SYS_LIBS)
+LIB_SYS_LIBS_Darwin += resolv
+LIB_SYS_LIBS_Linux += resolv
+LIB_SYS_LIBS_freebsd += resolv
+LIB_SYS_LIBS_solaris += resolv
 
 #===========================
 
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -5,8 +5,16 @@
  */
 
 #include <algorithm>
+#include <ctime>
+#include <cstring>
 #include <set>
 #include <tuple>
+
+#if defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) || defined(__sun)
+#  include <arpa/nameser.h>
+#  include <resolv.h>
+#  define PVXS_USE_DNS_TTL 1
+#endif
 
 #include <osiSock.h>
 #include <dbDefs.h>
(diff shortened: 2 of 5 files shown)


mdavidsaver:
What motivated use of the BSD `arpa/nameser.h` API vs. the `evdns` API?

On what thread is the (blocking?) DNS query performed?

derekste:
> What motivated use of the BSD `arpa/nameser.h` API vs. the `evdns` API?
> 
> On what thread is the (blocking?) DNS query performed?

The DNS lookup is currently blocking and can run on the PVXS tcp_loop thread during refresh. I kept this PR scoped to preserving the existing resolution path and adding lifetime refresh behavior, since moving refresh onto an async resolver path like evdns would be a significantly larger refactor. If we want to remove blocking DNS from the event-loop thread, that would be the next step in a follow-up change.

derekste:
Follow-up note on the last push:

The `f42fce6` change is only a build/link fix for the new DNS TTL code. The previous Makefile update added `resolv` through `LIB_SYS_LIBS_*`, but that did not reliably attach the resolver dependency to `libpvxs` itself in all CI configurations. That showed up in CI as unresolved resolver symbols (`ns_initparse` / `ns_parserr` in the static build, and `__res_query` in the Python wheel load test).

So this adjusts the dependency to `pvxs_SYS_LIBS_* += resolv` instead. That matches the usual EPICS target-specific pattern, makes the dependency follow the `pvxs` library directly, and keeps the change scoped to the library that actually uses the BSD resolver API. No functional/runtime behavior changed in this follow-up commit; it is just correcting how the new resolver symbols are linked.

derekste:
Superseding this DNS-first PR with a smaller reconfigure-focused branch based on the architecture discussion in https://github.com/epics-base/pvxs/issues/173. The new branch is https://github.com/derekste/pvxs/tree/dev/ifacemap-reconfigure and starts with IfaceMap-driven refresh of client UDP search destinations and server beacon destinations, leaving DNS/name-server/listener reconfiguration as follow-up work.

PR #175, opened 2026-05-01
Resolve pvxlist hostnames in server arguments

derekste:
## Summary

`pvxlist` server arguments are expected to work with `IP[:port]`, but hostnames/FQDNs can currently fail even when normal DNS resolution succeeds.

This resolves command-line server arguments before passing them to `.server()`, preserving the original argument in user-facing output.

## Example

Before:
pvxlist bidaq0-bbdcct.domain.gov
From bidaq0-bbdcct.domain.gov : Error resolving "bidaq0-bbdcct.domain.gov" : nodename nor servname provided, or not known

After
pvxlist bidaq0-bbdcct.domain.gov
BOOSTER:DCCTWF

also works for hostname(non-FQDN):port and other valid combinatons

### DIFF
--- a/tools/list.cpp
+++ b/tools/list.cpp
@@ -10,6 +10,7 @@
 #include <set>
 #include <list>
 #include <atomic>
+#include <cstring>
 
 #include <epicsVersion.h>
 #include <epicsGetopt.h>
@@ -57,6 +58,42 @@ void usage(const char* argv0)
             "  -w <sec>  Operation timeout in seconds.  Default 5 sec.  '0' disables timeout,\n"
             "            useful in combination with '-v'.\n"
             ;
+}
+
+std::string resolveServer(const char *arg, unsigned short defport)
+{
+    try {
+        return SockAddr(arg, defport).tostring();
+
+    }catch(...) {
+        auto original_error(std::current_exception());
+        const char *firstc = strchr(arg, ':'),
+                   *lastc  = strrchr(arg, ':');
+
+        if(firstc && firstc==lastc) {
+            in_addr addr;
+            std::string host(arg, firstc-arg);
+
+            if(!hostToIPAddr(host.c_str(), &addr)) {
+                SockAddr temp(AF_INET);
+                temp->in.sin_addr = addr;
+                temp.setPort(parseTo<uint64_t>(firstc+1));
+                return temp.tostring();
+            }
+
+        } else if(!firstc) {
+            in_addr addr;
+
+            if(!hostToIPAddr(arg, &addr)) {
+                SockAddr temp(AF_INET);
+                temp->in.sin_addr = addr;
+                temp.setPort(defport);
+                return temp.tostring();
+            }
+        }
+
+        std::rethrow_exception(original_error);
+    }
 }
 
 } // namespace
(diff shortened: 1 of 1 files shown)


mdavidsaver:
I am not able to reproduce the behavior described.

```
PVXS_LOG=*=DEBUG ./bin/linux-x86_64/pvxlist myhost
...
2026-05-14T15:53:56.309621723 DEBUG pvxs.client.io Connecting to 161.XXX.XXX.XXX:5075, RX readahead 262144
```

Name to address resolution is already done in `pvxs::SockAddr::setAddress()`

https://github.com/epics-base/pvxs/blob/3d854e492c44261ca25541dfda2177e934efb8b7/src/util.cpp#L523-L527

mdavidsaver:
I find myself in an uncomfortable position.  Not uncommon to FOSS maintainers these days, but new to me.  As a general rule I want to encourage new contributors.  Contributors make a healthy FOSS project.  However, it appears to me that an AI code generator created your PRs.  In the past, if I saw such c++ code I would assume it was produced by an inexperienced developer, and I would be inclined to make time to guide such a developer on the road towards experience.  This is how people learn, and how a developer community grows.  At least so far.

It is not clear to me how this impulse will translate to AI generated contributions.  How can I productively critique the code change of this PR?  I don't see how I can affect the LLM involved anymore than my published code corpus already has.  Should I ask you to please study the PVXS code base yourself?  Hoping that you will notice the redundancies introduced by, and poor placement of, these code changes?  In other words, that you will learn to detect and replicate the patterns which this LLM did not?

What I need to avoid is a situation where each contributor spends say ~10 minutes iterating on a PR, then I spend ~2 hours reviewing.  Then we repeat.  This does not scale up, and would quickly become unsustainable for me as a maintainer.  Should I compromise by providing less detailed reviews?  If I think you spent ~10 minutes generating, then I only spend ~10 minutes reviewing?  I fear that becomes a game of "bring me another rock".  A chatbot might not become frustrated, but the humans involved surely would.

I do not see any clear answers.  What should I do?

tboegi:
Some loose thougts here, trying to be helpful:
It looks as if @derekste had a problem with the host name resolver.
As it does not reprocuce for @mdavidsaver, we probably need to find out,
what happened.
What is different ? Network (of course) different DNS setups ? IPv4 vs IPv6 ?
We haven't come there yet.
Until that is clear, there is not much need to continue with a patch, I think.
The commit message should clearly:
- Have a headline (we have ?)
- describe the problem as exact as possible (I miss that)
- describe the choosen implementation, may be why it was choosen
   and not other implementations.
   (Here we may find a motivation for code duplication, kind of)
   
 The review, as I read it, indicates that the suggested solution is not ideal.
 For people who are not familiar with the code base, it may be helpful
 to point out where the original code is, that is now duplicated.
 This could help more people than just the one contributor.
 I assume that more people are reading here and not write anything.

So pointing out the code that should work as done here, hat is helpful. 
So yes, reviews are hard work.

derekste:
I over-focused on the symptom from my local environment and misdirected the fix into `pvxlist`, when the existing `.server()` path should already be resolving through `SockAddr::setAddress()`.

I’m going to revisit this patch and investigate why my host resolves via normal tools but fails through this path. I’ll gather resolver/debug output from the affected machine and come back with either a clearer reproducer or a narrower patch in the right layer. 👍

derekste:
I'll spare you the log dumps and cut to the chase: it was a local environment issue. I can't pinpoint the exact reason (we were recovering from an outage at the time), and I can't even say what version of pvxlist I was using (it has since been updated). Apologies for the noise. Withdrawn.

ralphlange:
Thanks, @derekste, for this quick and thorough reaction!
(Yes, @tboegi, there are more people reading.)

@mdavidsaver was scratching a related problem: With these tools becoming more widespread, we (as Base developers, but really as a collaboration) have to find a better way to organise code review. When creating code gets so much faster and easier, the create/review balance (that was hard to keep up anyway) gets whacked over.

I don't have a solution. Ideas very welcome! (But not on this ticket please - maybe in a discussion?)

Issue #176, opened 2026-05-12
QSRV PUT performance


mdavidsaver:
There is currently a large gap in performance between the run time of [`test/test1000.cpp`](test/test1000.cpp) and an equivalent [test](https://epics.anl.gov/tech-talk/2026/msg00562.php) with QSRV.

Some cursory profiling shows that PUT for both single and group PVs currently looks up the list of OS groups for every operation.  This need only be done once per connection.

mdavidsaver:
Frequent calls to `osdGetRoles()` dominate the cost of PUT for QSRV.

PR #178, opened 2026-05-17
QSRV PUT optimizations

mdavidsaver:
Attempting to address #176 by reducing calls to, and work by, `osdGetRoles()`.

- Restructure handling in single and group sources to make credential handling per Channel instead of per GET/PUT.
- Avoid some PUT specific allocations for GET.
- Add internal caching to `osdGetRoles()`.

Adds an expiring LRU cache to `osdGetRoles()` for targets using the `getpwnam()` or `NetUserGetLocalGroups()` implementations.  Cache is initially bounded to 128 entries, which will expire after 60 seconds.  This means that Channels created in the 60 seconds after a group list update becomes visible, would use the older list.  This has __security implications__.  I think acceptable to 99% of situations, where the group list is ~static.

@george-mcintyre fyi.

Python CI job failures due to https://github.com/epics-base/epicscorelibs/issues/48

### DIFF
--- a/ioc/groupsource.cpp
+++ b/ioc/groupsource.cpp
@@ -34,11 +34,28 @@ namespace ioc {
 
 DEFINE_LOGGER(_logname, "pvxs.ioc.group.source");
 
+/**
+ * group security cache - for storing group security credentials and clients
+ */
+class GroupSecurityCache : public SecurityControlObject {
+public:
+    // references entry in IOCGroupConfig::groupMap
+    Group& group;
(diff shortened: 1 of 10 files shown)


mdavidsaver:
Partially applied as 3c0154b413124bdcfa304b917f94d6bd53fbdf57.  Took the uncontentious parts relating to caching of Credentials by Channel.  Skip adding a global LRU cache to `osdGetRoles()` for now.

PR #179, opened 2026-05-22
qsrv & monitor: fix seven correctness issues

physwkim:
This series fixes seven independent correctness issues in QSRV (the IOC server), the access-security logging path, and the monitor pipeline. Each issue is a separate commit with its own regression test; every new assertion was confirmed to fail before its fix and pass after.

### Fixes

**1. `SecurityLogger` never restores `dbChannel::addr.pfield`** (`ioc/securitylogger.h`)
The data-carrying constructor saved `pDbChannel->addr.pfield` into `pfieldsave` but never set `pchan`, so both restore sites (constructor body and destructor), guarded by `if(pchan)`, were dead for every constructed logger. As the comment (and epics-base#474 / caPutLog#23) notes, `asTrapWrite` callbacks may clobber `pfield`; without the restore the saved pointer is dropped. Fixed by initializing `pchan(pDbChannel)` in the constructor init list, before the `asTrapWriteWithData` call.

**2. Non-atomic group put uses the wrong `SecurityClient` after a channel-less field** (`ioc/groupsource.cpp`)
`securityClients` is sized to `group.fields.size()` and indexed by the full field index, but the non-atomic put loop advanced `fieldIndex` only after a channel-backed field while `continue`-ing over channel-less `Structure`/`Const` fields (which sort first via the minimum putorder). After the first channel-less field the index drifts, so a write can be authorized/denied/logged with another field's security client. Fixed by advancing `fieldIndex` for every field and skipping only the DB op when `field.value` is null — matching the population, preprocessing, and atomic-put paths.

**3. Non-atomic group get omits `Const` fields** (`ioc/groupsource.cpp`)
The atomic get and monitor paths populate `Const` fields from `info.cval`, but the non-atomic get gated each read on `if(pDbChannel && leafNode)`, and `Const` fields have no channel by construction — so a `record[atomic=false]` get returned them at their default instead of the constant. Fixed by mirroring the atomic branch's field selection (skip only `Proc`/`Structure`, read the rest) and taking the per-field lock only when a channel is present.

**4. Monitor stats report the squash count in `nQueue`** (`src/servermon.cpp`)
`stats()` set `stat.nQueue = queue.size()` then immediately overwrote it with `nSquash`, and never set `stat.nSquash`. Fixed the second assignment to `stat.nSquash = mon->nSquash`.

**5. `ackAny` percentage is treated as a multiplier** (`src/servermon.cpp`, `src/clientmon.cpp`)
A percentage like `"50%"` was clamped to `[0, 100]` then used directly as `ackAt = percent * limit`, so 50% of a queue of 4 became 200 → clamped to the full queue; any percentage ≥ 1% collapsed to "ack at full queue". Fixed both parse paths to compute `percent / 100.0 * queueSize`. (The existing `ackAt == 0` guard and `max(1, …)` clamp still handle percentages too small to express on a short queue.)

**6. pvaLink never publishes the remote `timeStamp.userTag`** (`ioc/pvalink_lset.cpp`)
The monitor cache captures `timeStamp.userTag` into `fld_usertag`, and `snap_time` is updated from `fld_seconds`/`fld_nanoseconds`, but `snap_tag` (returned via `getTimeStampTag`) was never assigned from `fld_usertag` and stayed at its initial 0. Fixed by updating `snap_tag` from `fld_usertag` in the same snapshot block, with a 0 fallback.

**7. pvaLink type-change cleanup leaves `fld_seconds` stale** (`ioc/pvalink_link.cpp`)
`onTypeChange()` cleared the cached metadata fields but omitted `fld_seconds` (and assigned `fld_severity` twice). On a structure→non-structure type change, `pvaGetValue()` could keep using the old seconds, producing stale record timestamps. Fixed by clearing `fld_seconds` together with the other cached fields.

### Testing

Built against EPICS Base 7.0 (darwin-aarch64); each new test fails on `master` and passes with its commit.

- `testqsingle` — #1
- `testqgroup` — #2, #3
- `testmonpipe` — #4, #5
- `testpvalink` — #6, #7

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -27,9 +27,6 @@ jobs:
           skip: true
 
         - ml: manylinux2014_x86_64
-          py: cp38-cp38
-
-        - ml: manylinux2014_x86_64
           py: cp39-cp39
 
         - ml: manylinux2014_x86_64
@@ -43,9 +40,6 @@ jobs:
 
         - ml: manylinux2014_x86_64
           py: cp313-cp313
-
-        - ml: manylinux2014_x86_64
-          py: cp313-cp313t
 
         - ml: manylinux_2_28_x86_64
           py: cp314-cp314
--- a/ioc/credentials.h
+++ b/ioc/credentials.h
@@ -13,6 +13,7 @@
 #include <vector>
 #include <string>
 
+#include <pvxs/iochooks.h>
 #include <pvxs/source.h>
 
 namespace pvxs {
@@ -28,6 +29,7 @@ class Credentials {
 public:
     std::vector<std::string> cred;
     std::string host;
+    PVXS_IOC_API
     explicit Credentials(const server::ClientCredentials& clientCredentials);
     Credentials(const Credentials&) = delete;
     Credentials(Credentials&&) = default;
--- a/ioc/groupsource.cpp
+++ b/ioc/groupsource.cpp
@@ -464,17 +464,28 @@ void GroupSource::get(Group& group, const std::unique_ptr<server::ExecOp>& getOp
 
         // Loop through all fields
         for (auto& field: group.fields) {
-            dbChannel* pDbChannel = field.value;
+            // Read every non-Proc/Structure field, exactly as the atomic branch
+            // does.  Channel-backed fields are locked for the duration of their
+            // read; channel-less readable fields (e.g. MappingInfo::Const, which
+            // carry a constant value rather than a dbChannel) have no lock but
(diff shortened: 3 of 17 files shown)


mdavidsaver:
All 7 issues confirmed, and I noticed one more while reviewing (5353b02a33e4cd3d0e3c6de1caef3f1b2d4e292c).  I have applied much of this PR with modifications as 5b5ce4b28fa479ba25d3369d05ce749bc8c962ad through 7ae3ca84c085a2a5c90aa199d32aeb39fce9e340 inclusive.

PR #180, opened 2026-05-25
codec, client: bound decode and reject pre-INIT monitor data

physwkim:
Thanks for merging the earlier fixes. Three more I ran into while reading the
decode and client monitor paths, all remote-triggerable crashes. Numbers 1 and
2 are reachable during CONNECTION_VALIDATION, before auth.

**1. codec: bound value-decode recursion**
The depth guard in `from_wire()` is reset every time an `Any` *value* is decoded
(the inner type is read with a fresh descriptor), so it never bounds the
value-decode chain. A peer value that is a run of nested `Any` (one `0x82` byte
each) recurses one level per byte and overflows the stack. The SIGSEGV is not a
`std::exception`, so the dispatch try/catch misses it and the whole process
goes down. Fix threads a single `maxNestingDepth` across the `Any` boundary so
the bound holds for both type and value decode.

**2. codec: bound decode allocations to the remaining body**
An array length / struct child count is a peer-supplied 32-bit value, used to
allocate (POD) or default-construct every `Value` (StructA/UnionA/AnyA) before
the element bytes are read. A short frame claiming ~4.29e9 elements commits the
memory first, then faults on the short read. Fix caps the count against the
bytes still available (new `Buffer::maxAvail()`), since each element needs at
least one wire byte.

**3. client: reject a MONITOR data message before the INIT reply**
`handle_MONITOR()` decodes the payload before checking operation state, but
`info->fl` stays null until the INIT reply. A peer that has acked
CREATE_CHANNEL can send a data frame before INIT and the client dereferences
the null free-list. Fix moves the state check ahead of the decode.

Each has a regression test (testxcode, testclientconn). Built and run on
macOS/arm64.

### DIFF
--- a/src/clientmon.cpp
+++ b/src/clientmon.cpp
@@ -500,7 +500,47 @@ void Connection::handle_MONITOR()
                        peerName.c_str(), unsigned(ioid));
             return;
         }
+    }
 
+    // Validate the message against operation state *before* decoding any
+    // payload below.  The data branch dereferences info->fl, which stays null
+    // until the INIT reply has been processed, so a data message arriving while
+    // the monitor is still Creating must be rejected here -- not after the
+    // dereference.  Validating first makes "payload is decoded only after a
+    // successful INIT" hold by construction.
+
+    std::shared_ptr<OperationBase> op;
+    SubscriptionImpl* mon = nullptr;
+    if(M.good() && info) {
+        op = info->handle.lock();
+        if(!op) {
+            // assume op has already sent CMD_DESTROY_REQUEST
+            log_debug_printf(io, "Server %s ignoring stale cmd%02x ioid %u\n",
+                             peerName.c_str(), CMD_MONITOR, unsigned(ioid));
+            return;
+        }
+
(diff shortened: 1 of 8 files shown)


mdavidsaver:
I have applied parts 1 and 3 as 1679799691c9cf640bfefc8b277370dd9e7806a2 and 5c18e10ccf36c83c8d86db9edca9aa396e51ba1b.

I am still thinking about part 2.  PVA was not designed with untrusted peers in mind.  So all manner of remote resource exhaustion is possible.  I am not sure if mitigating this particular possibility is worth the added complexity.

physwkim:
Thanks for taking 1 and 3.

For 2, I see it a bit more like the same class of issue as 1 than a general resource-exhaustion problem. PVA's support for arbitrary nesting and flexible data structures naturally creates these kinds of edge cases where a very small input can drive disproportionately large work.

I don't think this patch meaningfully changes the overall threat model, but it does close off one of the more obvious and easily reachable cases. My impression from the TLS work is that there is some interest in improving robustness against less-trusted peers, and incrementally addressing these low-hanging cases seems consistent with that goal.

PR #184, opened 2026-05-27
Fix for PVA links to set UTAG/userTag

### LINKED ISSUE #177: PVA link with time=true doesn't copy userTag
When the PVA like is defined with time=true (and TSE-2) the content of `userTag` from the source is not copied to destination. 
I'm working on the pull request (with tests!)

https://github.com/epics-base/pvxs/blob/3d854e492c44261ca25541dfda2177e934efb8b7/ioc/pvalink_lset.cpp#L427-L429

joaopaulosm:
This is a proposal to fix #177 , I think I got it right.
Test case uses a source PV created from pvxs SharedPV with a customized userTag.

### DIFF
--- a/ioc/pvalink_lset.cpp
+++ b/ioc/pvalink_lset.cpp
@@ -4,6 +4,7 @@
  * in file LICENSE that is included with this distribution.
  */
 
+#include <cstdint>
 #include <epicsString.h>
 #include <alarm.h>
 #include <recGbl.h>
@@ -407,7 +408,8 @@ long pvaGetValue(DBLINK *plink, short dbrType, void *pbuffer, long *pnRequest) n
         }
 
         if(self->fld_usertag) {
-            self->snap_tag = self->fld_usertag.as<epicsUTag>();
+            // convert to uint32_t to avoid sign extension (Normative Types defines userTag as 32 bits)
+            self->snap_tag = self->fld_usertag.as<uint32_t>();
         } else {
             self->snap_tag = 0;
         }
--- a/ioc/pvalink_lset.cpp
+++ b/ioc/pvalink_lset.cpp
@@ -267,6 +267,9 @@ long pvaGetValue(DBLINK *plink, short dbrType, void *pbuffer, long *pnRequest) n
                                    self->pfieldname);
             if(self->time) {
                 plink->precord->time = self->snap_time;
+#ifdef DBR_UTAG
+                plink->precord->utag = self->snap_tag;
+#endif
             }
             log_debug_printf(_logger, "%s: %s not valid\n", __func__, self->channelName.c_str());
             return -1;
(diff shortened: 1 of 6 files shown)


ioc/pvalink_lset.cpp
```diff
@@ -267,6 +267,9 @@ long pvaGetValue(DBLINK *plink, short dbrType, void *pbuffer, long *pnRequest) n
                                    self->pfieldname);
             if(self->time) {
                 plink->precord->time = self->snap_time;
+#ifdef DBR_UTAG
```
physwkim: `#ifdef DBR_UTAG` is probably better as `#if DBR_UTAG`. `DBR_UTAG` is always
  defined (`0x20` on new Base, or `0` from the fallback on old Base), so
  `#ifdef` is always true and would break the build on old Base where there's no
  `utag` field. The rest of the code already uses `#if DBR_UTAG`
  (e.g. `iocsource.cpp:243`).
  
```suggestion
#if DBR_UTAG
```

physwkim (commented):
One more thing, maybe for a separate commit: `timeStamp.userTag` is a signed
`Int32`, so reading it back with `as<epicsUTag>()` sign-extends anything with
bit 31 set (`0x90000000` comes back as `0xFFFFFFFF90000000`). A counting
pulse-ID hits that all the time. Reading it as `uint32_t` avoids it.


```diff
// ioc/pvalink_lset.cpp  (read, around line 407)
-            self->snap_tag = self->fld_usertag.as<epicsUTag>();
+            self->snap_tag = self->fld_usertag.as<uint32_t>();
```

ioc/pvalink_lset.cpp
simon-ess: That seems to be the correct implementation here, though it seems that there are a few other spots in this code that have the same issue...

joaopaulosm: Good point, I'm fixing this for all other occurrences of `#ifdef DBR_UTAG`

joaopaulosm:
Good observation @physwkim, many thanks.
I modified accordingly to your suggestion and also changed the test to capture this scenario (2b12baa64d9c23ab0c2c02700ddb95e6a1691f09)
We wait until userTag will become 64 bits =)

mdavidsaver:
I have taken the liberty of pushing an rebase, and added a commit to fix the signed comparison failure on the `-Werror` CI job.  The synchronization failure on the RTEMS5 job needs investigation.  I think I am able to replicate on Linux.

mdavidsaver:
> I think I am able to replicate on Linux.

Nope.  An unrelated issue.

```
  Bailout called.  Further testing stopped:  ~QSrvWaitForLinkUpdate("target:utag") timeout at 2
  FAILED--Further testing stopped: ~QSrvWaitForLinkUpdate("target:utag") timeout at 2
```

I'm hoping this is just an artifact of timing.  Adding a new PV does not currently trigger a beacon anomaly, so the link client might have backed off long enough for the 10 second timeout in `testqsrvWaitForLinkConnected()` to expire.

mdavidsaver:
Applied with modifications as 26961de93110ed42d3b0affb4ccbb71caecb30a2.

The issue which appeared with the RTEMS5 job was due to incorrect usage of the (undocumented) `QSrvWaitForLinkUpdate` utility.  That object needs to be constructed before the `post()`.

https://github.com/epics-base/pvxs/blob/26961de93110ed42d3b0affb4ccbb71caecb30a2/test/testpvalink.cpp#L678-L680

Attempting to document...

https://github.com/epics-base/pvxs/blob/26961de93110ed42d3b0affb4ccbb71caecb30a2/ioc/qsrvpvt.h#L55-L65

PR #189, opened 2026-06-30
ioc: fix timestamp masking for monitor subscriptions

Monarda:
We don't use the `MappingInfo.nsecMask` at ISIS, and don't currently have any plans to do so.  As such, I may have misunderstood the reason for the difference between get and monitor behaviour.

In `subscriptionCallback` in `singlesource.cpp` a default-constructed `MappingInfo()` is passed to `IOCSource::get()`, leaving `nsecMask=0`.  Records with `info(Q:time:tag, "nsec:lsb:N")` had their nanoseconds field unmasked and userTag set incorrectly on monitor updates, while GET returned the right values (with lower bits masked).

This is a one line proposed fix and a regression test for the fix.

### DIFF
--- a/ioc/singlesource.cpp
+++ b/ioc/singlesource.cpp
@@ -55,8 +55,7 @@ void subscriptionCallback(SingleSourceSubscriptionCtx* subscriptionContext,
 
         {
             DBLocker F(dbChannelRecord(subscriptionContext->info->chan));
-            // TODO MappingInfo::nsecMask
-            IOCSource::get(currentValue, MappingInfo(), Value(), change, pChannel, pDbFieldLog);
+            IOCSource::get(currentValue, *subscriptionContext->info, Value(), change, pChannel, pDbFieldLog);
         }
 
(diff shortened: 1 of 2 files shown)


ioc/singlesource.cpp
```diff
@@ -55,8 +55,7 @@ void subscriptionCallback(SingleSourceSubscriptionCtx* subscriptionContext,
 
         {
             DBLocker F(dbChannelRecord(subscriptionContext->info->chan));
-            // TODO MappingInfo::nsecMask
-            IOCSource::get(currentValue, MappingInfo(), Value(), change, pChannel, pDbFieldLog);
+            IOCSource::get(currentValue, *subscriptionContext->info, Value(), change, pChannel, pDbFieldLog);
```
mdavidsaver, line 58: Right, well... clearly I forgot about this TODO.

mdavidsaver:
Applied as b5024df8fd7b05bc94b9fa8d1fbb7f1e0a9e2972.

PR #196, opened 2026-07-17
ioc: serve display.precision independent of DBR_GR_DOUBLE

physwkim:
## Problem

`getProperties()` (`ioc/iocsource.cpp`) nests the `display.precision`
assignment inside the `if(options & DBR_GR_DOUBLE)` branch:

```cpp
if(options & DBR_GR_DOUBLE) {
    dlL = meta.lower_disp_limit;
    node["display.limitHigh"] = meta.upper_disp_limit;
    if(options & DBR_PRECISION) {
        node["display.precision"] = int32_t(meta.precision.dp);
    }
}
```

`DBR_PRECISION` (`get_precision`) and `DBR_GR_DOUBLE` (`get_graphic_double`)
are **independent** rset slots. A field whose record supplies `get_precision`
but NULLs `get_graphic_double` therefore never has its precision served over
PVA — even though CA serves it fine. `display.units`, one block above, is
already gated on its own `DBR_UNITS` slot; `display.precision` was the odd
one out.

In EPICS base, `bo`, `mbbiDirect` and `mbboDirect` are exactly the record
types whose rset supplies `get_precision` while `#define get_graphic_double
NULL`; several modules (`busy`, `asyn`, `transform`, `sseq`) do the same. For
all of them the precision leaf is silently dropped over PVA.

## Fix

Gate `display.precision` on its own `DBR_PRECISION` slot, as a sibling of the
graphic-limits branch rather than nested inside it. One assignment moved up
one brace level.

## Test

Adds a `testqsingle` case using `test:precprop.HIGH` (`bo.HIGH` reports
`boHIGHprecision` = 2, while `bo` NULLs `get_graphic_double`). After the fix
its monitor delta carries `display.precision int32_t = 2` while
`display.limitLow`/`display.limitHigh` stay absent. The case **fails without
this fix** and passes with it. Full `testqsingle` run: 114/114.

## How it was found

Surfaced by a differential-oracle parity check of an independent Rust QSRV2
reimplementation against `softIocPVX` as ground truth: the read/monitor
marking diff resolved to exactly this family — 7 record types, 62 read
channels — every one a `display.precision`-only difference whose served value
matches what CA already reports. Only `bo.HIGH` carries a non-default
precision (2); the rest serve 0 (`recGblGetPrec` default), which is why the
drop had gone unnoticed.

### DIFF
--- a/documentation/releasenotes.rst
+++ b/documentation/releasenotes.rst
@@ -11,6 +11,8 @@ Release Notes
   "->booleanValue" may be replaced with "booleanValue".
 - tools: The CLI tools are now built only with epics-base >= 3.15 .
 - tools: pvxput now understand JSON syntax.
+- ioc: serve ``display.precision`` for a field whose rset supplies
+  ``get_precision`` but NULLs ``get_graphic_double`` (Sang Woo Kim)
 
 1.5.2 (Jun 2026)
 ----------------
--- a/ioc/iocsource.cpp
+++ b/ioc/iocsource.cpp
@@ -288,9 +288,14 @@ void getProperties(dbChannel* pChannel, db_field_log *pfl, Value& node)
         if(options & DBR_GR_DOUBLE) {
             dlL = meta.lower_disp_limit;
             node["display.limitHigh"] = meta.upper_disp_limit;
-            if(options & DBR_PRECISION) {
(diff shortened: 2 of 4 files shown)


mdavidsaver:
Applied as 8d9455a19d425043d22b66afd2ca3328a9929a6f with edits.  Thanks.

Issue #131, opened 2025-09-17
pvxs for RTEMS?


hjunkes:
Can I build pvxs for RTEMS? I'm missing libevent2 there.

mdavidsaver:
Yes.  RTEMS >=5 is supported.  There is a GHA job:

https://github.com/epics-base/pvxs/blob/021bcb4a06224ed8c0e8184b3a89c015830e2e3b/.github/workflows/ci-scripts-build.yml#L83-L88

I think a recipe to have rtems-source-builder build libevent would be a useful addition.  Until then, you can build the bundled libevent as described in my [documentation](https://epics-base.github.io/pvxs/building.html).

```
git clone --recursive https://github.com/epics-base/pvxs.git
...
make -C pvxs/bundle libevent.RTEMS-<yourbsp>
```

hjunkes:
According to https://docs.rtems.org/docs/main/user/migration/v5-to-v6.html I deleted '-spec bsp_specs' out of bundle/cmake/Platform/RTEMS.cmake

run
```
make -C pvxs/bundle libevent
make -C pvxs/bundle libevent.RTEMS-qoriq_e500
make -C pvxs
```
without problems:
```
rtems@rtems-dev:~/MVME2500_6/EPICS/epics-support$ ls -l pvxs/lib/RTEMS-qoriq_e500/
total 32088
-rw-r--r-- 1 rtems rtems 24404116 Sep 18 12:12 libpvxs.a
-rw-r--r-- 1 rtems rtems  8449574 Sep 18 12:14 libpvxsIoc.a
```

hjunkes:
Hallo Michael,
I am now trying to get the whole thing running with RTEMS6 / libbsd on an MVME2500.
Unfortunately, the whole thing hangs in the call UDPManager::instance at 
pvxs/src/server.cpp 
```
Server::Pvt::Pvt(const Config &conf)
  384     :effective(conf)
  385     ,beaconMsg(128)
  386     ,acceptor_loop("PVXTCP", epicsThreadPriorityCAServerLow-2)
  387     ,beaconSender4(AF_INET, SOCK_DGRAM, 0)
  388     ,beaconSender6(AF_INET6, SOCK_DGRAM, 0)
  389     ,beaconTimer(__FILE__, __LINE__,
  390                  event_new(acceptor_loop.base, -1, EV_TIMEOUT, doBeaconsS, this))
  391     ,searchReply(0x10000)
[...]

PR #197, opened 2026-07-27
rtems: gate the RTEMS 5 workarounds that break RTEMS 6

physwkim:
Both are RTEMS 5-era workarounds that break RTEMS 6 — `-specs bsp_specs` (removed in RTEMS 6) blocks the build, and avoiding kqueue puts libevent on the poll backend where libbsd's POLLERR on its notify pipe spins `event_base_loop()` before `iocInit` — so gate each on `CMAKE_SYSTEM_VERSION` / `__RTEMS_MAJOR__` rather than remove it.

### DIFF
--- a/bundle/cmake/Platform/RTEMS.cmake
+++ b/bundle/cmake/Platform/RTEMS.cmake
@@ -24,8 +24,16 @@ set(CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES
 )
 set(CMAKE_SYSTEM_LIBRARY_PATH ${CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES})
 
+# RTEMS 6 no longer installs a BSP spec file.  Passing -specs bsp_specs to
+# an RTEMS 6 toolchain fails before anything is compiled.
+if(CMAKE_SYSTEM_VERSION VERSION_LESS 6)
+  set(_rtems_bsp_specs "-specs bsp_specs ")
+else()
+  set(_rtems_bsp_specs "")
+endif()
+
 set(CMAKE_C_FLAGS_INIT
- "-B${RTEMS_TARGET_PREFIX}/${RTEMS_BSP}/lib/ -specs bsp_specs -qrtems ${RTEMS_BSP_C_FLAGS}"
+ "-B${RTEMS_TARGET_PREFIX}/${RTEMS_BSP}/lib/ ${_rtems_bsp_specs}-qrtems ${RTEMS_BSP_C_FLAGS}"
 )
 set(CMAKE_C_FLAGS_INIT ${CMAKE_C_FLAGS_INIT})
 
--- a/src/evhelper.cpp
+++ b/src/evhelper.cpp
@@ -11,6 +11,10 @@
 #  include <mswsock.h>
 #endif
 
+#ifdef __rtems__
+#  include <rtems/score/cpuopts.h>
+#endif
+
 #include <cstring>
 #include <system_error>
 #include <deque>
@@ -175,12 +179,20 @@ struct evbase::Pvt final : public epicsThreadRunable
         try {
             evconfig conf(__FILE__, __LINE__, event_config_new());
 #ifdef __rtems__
+#  if (__RTEMS_MAJOR__ < 6) || (__RTEMS_MAJOR__ == 6 && __RTEMS_MINOR__ <= 2)
             /* with libbsd circa RTEMS 5.1
              * TCP peer close/reset notifications appear to be lost.
              * Maybe due to absence of NOTE_EOF?
              * poll() seems to work though.
+             *
+             * The rtems-libbsd descriptor fixes restoring kqueue ship in
+             * RTEMS 6.3; the version condition follows the rtems-libbsd
+             * maintainer's recommendation.  Note the poll backend spins
+             * on RTEMS 6: libbsd reports POLLERR for the pipe() libevent
+             * notifies itself through.
              */
             event_config_avoid_method(conf.get(), "kqueue");
+#  endif
 #endif
             decltype (base) tbase(__FILE__, __LINE__, event_base_new_with_config(conf.get()));
             if(evthread_make_base_notifiable(tbase.get())) {


mdavidsaver:
Does this change relate to #131?  What specific RTEMS/libbsd version(s) and BSPs have been tested?

@hjunkes fyi.

hjunkes:
I’m not sure if this belongs here?
These are my changes to libevent for RTEMS (6)/7

[bundle_libevent.patch](https://github.com/user-attachments/files/30422464/bundle_libevent.patch)

physwkim:
I hadn't seen #131 when I opened this — but yes, it's the same two changes @hjunkes made by hand there, gated on the RTEMS major instead of removed, since `bsps/arm/xilinx-zynq/start/bsp_specs` still exists on the RTEMS 5 branch.

Tested on RTEMS 6.0.0 (kernel `2faafecb`) with rtems-libbsd `6-freebsd-12` @ `96319dbd`, BSP `arm/xilinx_zynq_a9_qemu` under qemu 8.2.2, arm-rtems6-gcc 13.3.0, EPICS 7.0.10, PVXS `cc7bc72`, bundled libevent `1fe626c4` unpatched. Stock `evhelper.cpp` there wedges at 98.4 % CPU and never reaches `iocInit`; gated, the IOC serves (`pvxget`/`pvxput` round trip, ~8 % CPU).

On RTEMS 5 I couldn't check — no arm-rtems5 toolchain here.

mdavidsaver:
> gated on the RTEMS major

Is this a question of RTEMS 5/6/7, or of libbsd vs. legacy network?

I'm not asking for broad compatibility with RTEMS/libbsd versions.  I just want to be clear about which combinations have been tested successfully, and which are known not to work.

physwkim:
RTEMS 5 vs 6+ (`__RTEMS_MAJOR__` / `CMAKE_SYSTEM_VERSION`); tested only RTEMS 6.0.0 + rtems-libbsd `6-freebsd-12` on `arm/xilinx_zynq_a9_qemu`, where kqueue serves and the poll backend wedges — RTEMS 5/7 and the legacy/lwIP stacks are untested by me.

physwkim:
@mdavidsaver — you were right to push on this, thanks. Measured on `arm/xilinx_zynq_a9_qemu`: on RTEMS 6 and 7 **neither** backend is usable as it stands. The poll build never reaches `iocInit` (94.0 % / 99.2 % host CPU, log byte-frozen), and the kqueue build boots but the suite degrades — RTEMS 6: testget 43 ok of 68, testmon 11 of 45, testput 18 of 52, testqgroup 27 of 46, testqsingle 59 of 113, testrpc takes a CPU exception in `PVXTCP`; RTEMS 7: most binaries still running when the 240 s cap fires. So the gate in this PR picks between two broken configurations on 6/7, and I should not have presented it as a fix.

The causes are all below pvxs, in rtems-libbsd and the RTEMS kernel. Seven reports:

- [rtems-libbsd#101](https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/work_items/101) — `dup()` / `fcntl(F_DUPFD)` answer `ENXIO` for every libbsd descriptor
- [rtems-libbsd#102](https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/work_items/102) — kqueue: the hold is leaked when `fget()` fails, and `closefp()` drops the wrong descriptor's knotes
- [rtems-libbsd#103](https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/work_items/103) — `select()` and `poll()` pass `rtems_bsd_libio_fo_poll()` three different descriptors
- [rtems-libbsd#104](https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/work_items/104) — `rtems_bsd_sysgen_read()` reads a vnode out of a descriptor that has none; this is the `EINVAL` from `read()` on libevent's notify socket
- [rtems-libbsd#39](https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/work_items/39#note_155673) (existing) — `pipe()` is the IMFS FIFO, so `poll()` answers `POLLERR`: the wedge above
- [rtems#5667](https://gitlab.rtems.org/rtems/rtos/rtems/-/work_items/5667) — `fcntl(F_DUPFD_CLOEXEC)` fails with `EINVAL`
- [rtems#5668](https://gitlab.rtems.org/rtems/rtos/rtems/-/work_items/5668) — `fcntl(F_DUPFD)` ignores its third argument (conformance; not needed for the suite)

With the six fixes applied (branches at https://github.com/physwkim/rtems-libbsd), all four combinations — RTEMS 6 and 7 × poll and kqueue — run the 29 pvxs test binaries identically, suite for suite, to the RTEMS 5 + poll column, and every column has zero `not ok` that is not a pvxs `# TODO`. testdiscover still stops at 2 of 8 everywhere, RTEMS 5 included, which is independent of this.

@kiwichris — could you check that the seven reports are valid? Four of them are in descriptor handling that libbsd's own testsuite does not exercise: nothing in `testsuite/` calls `pipe()`, `dup()` or `fcntl(F_DUPFD*)`.

kiwichris:
Thank you for the detailed analysis and the patches. I am digesting them. My initial review says they are valid. Duplicating of networking file descriptors has been a weakness ever since networking was added. It is good to see this level of focus on it.

I suggest we concentrate on `7-freebsd-14` and the kernel fixes for `main`. Once reviewed and merged we can back port them. Could you please create MRs on RTEMS GitLab?

Note, as stated in our GitLab issues `6-freebsd-12` is not being maintained and we recommend all users move to `6-freebsd-14`. That branch of code has too many other issues fixed in the FreeBSD 14 release.

physwkim:
Thank you for taking a look. The MRs are up:

- https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/merge_requests/153 (`7-freebsd-14`, five fixes with a testsuite case each)
- https://gitlab.rtems.org/rtems/rtos/rtems/-/merge_requests/1381 (kernel `main`, F_DUPFD_CLOEXEC + the F_DUPFD minimum, with psx13 cases)
- https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/merge_requests/154 (`6-freebsd-14` version of !153 — happy to hold this one until !153 has been reviewed and merged, per your sequencing)

hjunkes:
Together with Claude, I tested the changes and enhancements on actual hardware this morning (MVME6100, RTEMS 7). I then asked someone to write me a summary so that I wouldn’t forget anything, as I experienced a few crashes along the way that might have something to do with pipe()/pipe2()?

# MVME6100_7 rtems-libbsd/kernel update — session summary (2026-07-30)

Target: `rtems-dev.fhi.mpg.de:/home/rtems/MVME6100_7`, MVME6100 board "gonzo"
(powerpc/beatnik BSP, RTEMS 7), EPICS IOC `myExample`.

## What was integrated

**rtems-libbsd** (branch `7-freebsd-14`), merged from
`physwkim/rtems-libbsd` branch `fix/descriptor-identity-7` (fast-forward,
base commit matched exactly):

| commit | title | closes |
|---|---|---|
| `dfa66235` | rtemsbsd/libio: Give a duplicated descriptor an identity of its own | rtems-libbsd#101 |
| `92df493d` | rtemsbsd/kqueue: Release the hold taken on a registered descriptor | rtems-libbsd#102 |
| `0198e7f0` | rtems/syscall: Give pipe() and pipe2() the libbsd implementation | rtems-libbsd#39 |
| `ebe7749f` | freebsd/sys_generic: Pass the scan the descriptor whose hold it drops | rtems-libbsd#103 |
| `d4f32ebb` | rtems/syscall: Do not read a vnode out of a descriptor that has none | rtems-libbsd#104 |

Also committed, pre-existing local work (unrelated to the above, already
present uncommitted in the checkout before this session):

| commit | title |
|---|---|
| `1aa2a0a2` | rtemsbsd/dev/mve: Fix spurious-interrupt race during attach |

**RTEMS kernel** (`rtos/rtems`, branch `main`), cherry-picked from
`physwkim/rtems` MR `!1381` (`fix/fcntl-dupfd`, not yet merged upstream):

| commit | title | closes |
|---|---|---|
| `a6dea3bd93` | libcsupport: Add fcntl(F_DUPFD_CLOEXEC) | rtems#5667 |
| `dfc5c8b132` | libcsupport: Honour the file descriptor minimum of fcntl(F_DUPFD) | rtems#5668 |

The other two commits in that MR — `539cc24a` (fcntl(F_DUP2FD) onto a free
descriptor, closes #5670) and `80092e7a` (dup2() onto a non-open descriptor,
closes #5669) — were deliberately **not** cherry-picked; only #5667/#5668
were requested this session.

Both repos rebuilt (`./waf` / kernel `./waf install`) and reinstalled
cleanly; `myExample` IOC rebuilt and relinked against the new kernel +
libbsd without issue.

## Crash on external pvAccess monitor

After deployment, an external `pvmonitor rtems:line` (both a Mac-built
EPICS-Base `pvmonitor` and, later, pvxs's own `pvxmonitor` tool) reliably
crashed the IOC on the **first** boot after deployment:

```
*** FATAL ***
fatal source: 9 (RTEMS_FATAL_SOURCE_EXCEPTION)
exception vector 3 (0x3)          <- PowerPC DSI (data storage exception)
DAR = 0x7c000278                  <- equals R30
```

`DAR`/`R30` (`0x7c000278`) decodes as a plausible **PowerPC instruction
word**, not a heap address — a strong signal of reading a corrupted or
never-initialized pointer rather than a null/wild jump.

Symbolizing the fault PC and stack trace against the IOC binary showed the
crash inside **pvxs's own event-loop internals**, not directly in any of
the patched libbsd/kernel code:

```
std::unique_ptr<pvxs::impl::mdetail::VFunctor0>::~unique_ptr()
pvxs::impl::mfunction::~mfunction()                      (evhelper.h)
pvxs::impl::evbase::Pvt::Work::~Work()                    (evhelper.cpp)
std::deque<Work>::_M_move_assign1(...)                     <- inside doWork()
pvxs::impl::evbase::Pvt::doWork()
pvxs::impl::evbase::Pvt::doWorkS()
event_process_active / event_base_loop
pvxs::impl::evbase::Pvt::run()                             <- the event-loop worker thread
```

I.e. a queued work-item closure (`mfunction`, wrapping a
`unique_ptr<VFunctor0>`) is destroyed holding a pointer that looks like raw
code bytes rather than a valid heap address.

### Bisection

Reverting `0198e7f0` alone (pipe()/pipe2() → back to the broken IMFS FIFO)
made the crash disappear immediately and repeatably; restoring it brought
the crash back. This isolates the trigger cleanly to that one commit.

### What was ruled out

- `chgpipecnt()` touching an uninitialized/stub RTEMS credential
  (`cr_ruidinfo`) in `sys_pipe.c`'s `pipespace_new()` — all four call sites
  are correctly `#ifndef __rtems__`-guarded in the shipped code; not
  reachable on RTEMS.
- A bad/stale build — the exact same source tree was rebuilt from scratch
  and redeployed **three separate times** after the first crash; it never
  crashed again on any subsequent boot, including with the same client
  that originally triggered it. A stale-object-file explanation would be
  expected to fail consistently, not once-then-never-again.

### Leading (unconfirmed) hypothesis

pvxs bundles its own libevent with a local RTEMS patch
(`bundle/libevent/evutil.c`, pre-existing local diff, not part of this
session's changes):

```c
#if defined(EVENT__HAVE_PIPE) && !defined(__rtems__)
    if (pipe(fd) == 0) { ... }
#endif
```

i.e. on RTEMS, `evutil_make_internal_pipe_()` (libevent's cross-thread
event-loop wakeup mechanism) has **no fallback** — it depends entirely on
`pipe2(fd, O_NONBLOCK|O_CLOEXEC)` succeeding. Before `0198e7f0`, `pipe2()`
silently resolved to the broken IMFS FIFO (or failed outright); after it,
this is a real, non-blocking, cross-thread-accessed BSD pipe for the first
time. This is exactly the code path newly exercised by the fix, and is the
most likely place for a genuine bug, but it was not confirmed.

### Timing sensitivity (heisenbug behavior)

- Enabling full syscall/descriptor tracing (`RTEMS_BSD_SYSCALL_TRACE` /
  `RTEMS_BSD_DESCRIP_TRACE`) did not reproduce the crash — it instead
  caused an unrelated hang on `recvfrom()` during early boot/DHCP, clearly
  from the console-I/O overhead perturbing timing.
- Light, targeted instrumentation (a `printk()` on every `pipe()`/`pipe2()`
  call, plus one on every `mfunction` construction/destruction, logging the
  wrapped raw pointer) compiled cleanly and ran through **hundreds of
  clean dispatch cycles** with no corruption, across two separate live
  tests (a 90s sustained `pvxmonitor` run, and a live rerun of the original
  crashing `pvmonitor` client) — zero reproductions.

Both observations point to a genuine, narrow-window race rather than a
deterministic logic bug, most likely in `sys_pipe.c`'s non-blocking,
concurrent-thread read/write path as exercised by libevent's notify pipe.
**This was not root-caused.** All diagnostic instrumentation was reverted
before finishing; the final deployed build is clean production code with
all five libbsd fixes and both kernel fixes active.

### Unrelated finding en route (also unconfirmed)

While investigating, we noticed the pre-existing `mve_isr()` guard (commit
`1aa2a0a2`, not part of physwkim's work) returns early without calling
`BSP_mve_disable_irqs()` when `sc->mp`/`sc->daemonTid` aren't yet valid —
since `sc->mp` is NULL in that branch, the hardware IRQ can't be
acknowledged there. We hit one "Discovery: Spurious interrupt" storm during
live debugging (most plausibly triggered by an unrelated raw/malformed
telnet connection to the board's telnetd) that required a physical
power-cycle to clear. We couldn't conclusively tie it to this guard (the
BSP's spurious-interrupt print path doesn't exactly match what the guard
would cause), so it's flagged as a separate, real-looking defect worth a
second look, not confirmed as the cause of that specific incident.

## kqueue end-to-end validation

Separately from the crash investigation: pvxs itself unconditionally
disables the kqueue backend on RTEMS (pre-existing code, `evhelper.cpp`):

```c
#ifdef __rtems__
    /* with libbsd circa RTEMS 5.1 TCP peer close/reset notifications
     * appear to be lost. poll() seems to work though. */
    event_config_avoid_method(conf.get(), "kqueue");
#endif
```

So none of this session's kqueue-related fixes (`92df493d` / #102, and
`0198e7f0` / #39, which kqueue's `EV_ADD` also depends on) were actually
being exercised by this IOC at all — it always ran on poll().

We made the avoidance conditional on a new `RTEMS_ALLOW_KQUEUE` env var
(default unset = old poll-only behavior, so this is fully backward
compatible and revertible via `st.cmd` alone, no rebuild needed) and
enabled it. Confirmed via log output that kqueue was genuinely selected:

```
INFO pvxs.loop Enter loop worker for 0x1962b60 using kqueue
```

`pvmonitor rtems:line` then ran cleanly under the kqueue backend, streaming
updates with no crash — end-to-end confirmation that #102 and #39 work
correctly through the real application, not just the lower-level pvxs
test-suite binaries.

## Current deployed state

- Kernel `main`, 2 commits ahead of `origin/main` (fcntl fixes).
- rtems-libbsd `7-freebsd-14`, 6 commits ahead of `origin/7-freebsd-14`
  (5 descriptor-identity fixes + the local mve driver fix).
- `myExample` IOC boot image built clean (no diagnostics), currently
  running with `RTEMS_ALLOW_KQUEUE=1` set in `st.cmd` (kqueue active).
  Reverting to poll is a one-line `st.cmd` edit + reboot.

## Open questions for Sang-Woo Kim

1. Any suspicion of a concurrency bug in the RTEMS port of `sys_pipe.c`
   under non-blocking, cross-thread read/write access (as used by
   libevent's self-pipe notify trick)? That's our leading hypothesis for
   the crash, but we couldn't reproduce it under instrumentation to
   confirm.
2. Does the `mve_isr()` early-return guard (unrelated local driver code,
   not yours) needing to still acknowledge the hardware IRQ even when
   `sc->mp`/`daemonTid` aren't ready seem like a real concern from your
   experience with this class of bug, or a red herring?
3. Since the crash never reproduced under any instrumentation, is there a
   lower-overhead way (e.g. a lightweight event/counter you'd trust more
   than `printk` to a slow UART) you'd recommend for chasing a
   narrow-window race like this on RTEMS?

physwkim:
Thanks for the thorough testing, @hjunkes. I could not reproduce it either, on an RTEMS 7 QEMU rig following your exact path — poll backend, libevent self-pipe on pipe2 (`EVENT__HAVE_PIPE2`), external pvxmonitor first-connect after fresh boot ×8, plus connect/disconnect cycling and a sustained monitor — all clean.

One related thing came out of the libbsd review: cpukit and rtems-libbsd now both define `pipe()`, and which one lands in an image is decided only by archive order on the link line (see the discussion on [rtems-libbsd!153](https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/merge_requests/153); now [rtems!1382](https://gitlab.rtems.org/rtems/rtos/rtems/-/merge_requests/1382)). That cannot corrupt a single consistently built image, but it is one way an incrementally or partially rebuilt image can silently differ from a clean one — which fits the pattern you saw: one crash on a freshly deployed image, bisect flips implicating the pipe commit, and no reproduction after clean rebuilds.

It does not mechanically explain the DSI signature you captured, so if it ever comes back on a cleanly built image, the ELF and boot log would be very useful. Picking up rtems!1382 just removes the link-order dependence going forward.

physwkim:
> Any suspicion of a concurrency bug in the RTEMS port of sys_pipe.c under non-blocking, cross-thread read/write access (as used by libevent's self-pipe notify trick)? That's our leading hypothesis for the crash, but we couldn't reproduce it under instrumentation to confirm.

I couldn't find anything — the port keeps FreeBSD's `sys_pipe.c` locking unchanged, and ~61M cross-thread nonblocking notifies with heap canaries (plus the full IOC scenario) stayed clean. I'd also expect a pipe race to damage the pipe's own buffer rather than an unrelated heap object — but I may be missing something.

> Does the mve_isr() early-return guard (unrelated local driver code, not yours) needing to still acknowledge the hardware IRQ even when sc->mp/daemonTid aren't ready seem like a real concern from your experience with this class of bug, or a red herring?

I don't know the driver, but a missing ack usually shows up as an interrupt storm rather than a one-shot scribble. What I'd want to rule out nearby: MOTLoad has just used the NIC, so DMA state left un-quiesced at attach could write into what is now heap — which would fit "once on a fresh image, never again".

> Since the crash never reproduced under any instrumentation, is there a lower-overhead way (e.g. a lightweight event/counter you'd trust more than printk to a slow UART) you'd recommend for chasing a narrow-window race like this on RTEMS?

`rtems_record`, or a static ring of {timebase, site, value} dumped from a fatal extension — only stores on the hot path, UART only after the crash. If a crash yields a corrupted address again, a DABR watchpoint on it could catch the writer directly.

physwkim:
> While investigating, we noticed the pre-existing mve_isr() guard (commit 1aa2a0a2, not part of physwkim's work) returns early without calling BSP_mve_disable_irqs() when sc->mp/sc->daemonTid aren't yet valid

One data point that may help: I could not find commit 1aa2a0a2 in upstream rtems-libbsd (no ref has it after a fetch), and the stock `mve_isr` there has no guard — it unconditionally calls `BSP_mve_disable_irqs(sc->mp)` and posts the event. So the guard looks local to your tree.

Both variants share the same underlying window, though: if the IRQ can fire while `sc->mp` is still NULL, the stock code NULL-derefs and the guarded one returns with the level source still asserted — a storm, which fits the incident you saw (crash vs. hang, same window). What would close both is ordering at attach: quiesce the controller first (stop DMA, mask, clear pending — MOTLoad has just been using it), then build the softc, and only then connect and enable the ISR. If a defensive guard is kept, disabling the vector at the interrupt-controller level (which does not need sc->mp) before returning would at least prevent the storm.

kiwichris:
>     * https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/merge_requests/153 (`7-freebsd-14`, five fixes with a testsuite case each)

Merged. Thanks.

>     * https://gitlab.rtems.org/rtems/rtos/rtems/-/merge_requests/1381 (kernel `main`, F_DUPFD_CLOEXEC + the F_DUPFD minimum, with psx13 cases)

Under review.

I have raised https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/merge_requests/153 to deal with the `pipe` and `pipe2` calls in the kernel. Once merged the `7-freebsd-14` branch can be updated to register handlers.

physwkim:
The mve_isr() argument bug was already reported by @hjunkes as [rtems-libbsd#94](https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/work_items/94); the attach-window race is now [rtems-libbsd#105](https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/work_items/105), and [rtems-libbsd!155](https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/merge_requests/155) fixes both. @hjunkes, once the pipe/pipe2 rework @kiwichris raised ([rtems!1383](https://gitlab.rtems.org/rtems/rtos/rtems/-/merge_requests/1383)) has landed, could you try !155 on a tree without your local mve patches?

hjunkes:
Thanks for the detailed answers — very helpful. Update from this morning's testing:

Tested !155 on a tree without the local mve patch, as requested: checked out !155 (base + the 5 rebased 
descriptor-identity fixes + your two mve commits, no local 1aa2a0a2), rebuilt/reinstalled rtems-libbsd,
rebuilt/redeployed the IOC, rebooted, and ran an external pvmonitor against it.

The crash reproduced, identically — DAR/R30 = 0x7c000278, the exact same value as the original crash. Same deterministic-looking corruption pattern, happening whether or not #94/#105 are fixed. So this confirms the crash is unrelated to mve_isr, and !155 itself is validated as correct — we've now adopted it as our baseline, replacing the local guard-based patch.

Instrumented the suspected code path with a lightweight, fatal-safe ring buffer instead of printk (which had already twice failed to reproduce the crash under observation): a static, lock-free, I/O-free ring of {timebase, site, ptr, ptr} entries, written from mfunction's constructor/destructor with no locks/allocation/console touch on the hot path (rtems_counter_read() for timing), dumped via printk only from an RTEMS fatal extension (rtems_extension_create()) if
  a fatal exception actually occurs — so it's silent during normal operation.

Result: three clean runs in a row under this instrumentation (a 90s automated pvxmonitor session, and two separate pvmonitor runs from the original crashing client) - zero reproductions. Same heisenbug signature as before: every crash we've actually seen happened with no instrumentation present; every instrumented run, even this near-zero-overhead one, has been clean. Strongly suggests a very narrow, binary-layout/cache-sensitive race window.

Rather than keep forcing reproduction attempts, we've left the instrumented build deployed under normal IOC operation - it's cheap and dormant, so if the crash recurs naturally the fatal extension will dump the last ~500 mfunction events automatically.

We're also now bringing up a second, independent target (real VME hardware, a Highland V375 AWG, more cards to follow) that will start exercising pvxs as well, so there's a second, differently-loaded system that may help narrow this down further. Will report back if either system catches anything.

physwkim:
Just to confirm, was RTEMS MR !1383 included in this test configuration as well? I couldn’t tell from the description.

physwkim:
@hjunkes Do you run `waitProbe; netShut` at the MOTLoad prompt before booting RTEMS? If not, that may be the story: nothing in software quiesces the GbE ports before `rtems_bsd_initialize()`, and 0x7c000278 matches the RX descriptor write-back word the port would DMA into a MOTLoad-armed ring (OWN cleared, FIRST|LAST set, INT_ENA preserved) — a fixed-address scribble that would also explain why any instrumentation (heap layout shift) hides the crash.

hjunkes:
target gets bootet with this sequence, mot-script-boot:
```
MVME6100> gevShow
mot-script-boot
tftpGet -a04000000 -c141.14.128.12 -s141.14.131.192 -m255.255.240.0 -g141.14.128.128 -frtems6/myExample.boot
waitProbe
netShut
go -a04000000
```

Good catch, no — !1383 was not included in this test configuration. Our RTEMS kernel tree is origin/main plus only two unrelated local commits (fcntl F_DUPFD/F_DUPFD_CLOEXEC fixes, closing #5667/#5668); the pipe/pipe2 indirect-handler change from !1383 isn't present anywhere in our history.
So the !155 test result stands independent of that MR — worth keeping in mind if !1383 is itself a candidate fix for the crash, since we haven't exercised it yet. Happy to pull it in and retest if you think it's relevant to the self-pipe race.

physwkim:
@mdavidsaver @hjunkes A direction question: now that the kqueue fixes are merged in rtems-libbsd 7-freebsd-14 ([!153](https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/merge_requests/153); the 6-freebsd-14 backport [!154](https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/merge_requests/154) pending), should pvxs keep avoiding kqueue on RTEMS (`event_config_avoid_method(conf, "kqueue")` in evhelper.cpp)? The avoidance dates to libbsd circa RTEMS 5.1 losing TCP close/reset notifications ([rtems-libbsd#69](https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/issues/69)), which those fixes address. Defaulting to kqueue would drop the self-pipe entirely (EVFILT_USER wakeups — no pipe()/pipe2() dependency) and poll's per-call per-fd allocation churn in libbsd; our RTEMS 6/7 QEMU targets run the full pvxs ladder with kqueue enabled. Thoughts on the direction?

hjunkes:
!1383 wasn't in our tree yet (GitLab API showed it still opened, not merged, at test time). Pulled it directly
(refs/merge-requests/1383/head), cherry-picked the single commit onto our kernel main (applied cleanly, no conflicts),
rebuilt the kernel BSP and rtems-libbsd, and redeployed. One caveat: !1383 only adds the pipe/pipe2 indirect-handler infrastructure — nothing in our current rtems-libbsd registers an alternate handler through it yet, so pipe()/pipe2() still resolve to the identical default code path. It's inert for us as-is; is there a companion rtems-libbsd change (existing or planned) that would register a BSD-native pipe through this hook? That's the piece that would actually bear on the self-pipe/libevent race.

Rebooted and re-ran pvmonitor against the !1383-included build anyway — another clean run, no reproduction (fourth in a row under instrumentation now). Also added pvxs/qsrv support to our second target (beatnik, the real-VME-hardware system) so it's now exercising the same event-loop path too — no results yet, will report if it catches anything.

@physwkim: Regarding the question of whether kqueue should be the default. We have an RTEMS-EPICS meeting tomorrow evening. I’ll raise this there. I think kqueue should be enabled by default. RTEMS5 is no longer supported with EPICS7.

physwkim:
@hjunkes Thanks for the follow-up and the testing. Yes, that's the plan once !1383 merges: drop the strong pipe()/pipe2() from rtems-bsd-syscall-api.c and register the BSD implementation through the new hooks from a SYSINIT. To set expectations: pipe2() already resolves to libbsd today, so this removes the link-order fragility rather than changing behavior.

mdavidsaver:
> ... should pvxs keep avoiding kqueue on RTEMS ...

My preference is for PVXS to detect the fixed version and adjust accordingly.  Either at compile or runtime.  eg. a version macro to compare.  Or a specific feature flag macro.

@kiwichris Can you recommend a condition which can be used?

kiwichris:
> > ... should pvxs keep avoiding kqueue on RTEMS ...
> 
> My preference is for PVXS to detect the fixed version and adjust accordingly. Either at compile or runtime. eg. a version macro to compare. Or a specific feature flag macro.
> 
> @kiwichris Can you recommend a condition which can be used?

The LibBSD fixes recently merged and the one pending fix for the duplicated `pipe` and `pipe2` calls will be present in RTEMS 6.3 and later releases including RTEMS 7. Something like this should work:
```
#if (__RTEMS_MAJOR__ > 6) || (__RTEMS_MAJOR__ == 6 && __RTEMS_MINOR__ > 2)
```

physwkim:
Updated as suggested: kqueue is now enabled from RTEMS 6.3 using the condition @kiwichris recommended; earlier releases keep the avoidance.

physwkim:
rtems!1383 has been merged; the libbsd follow-up registering the pipe handlers through the new cpukit hooks is now filed as https://gitlab.rtems.org/rtems/pkg/rtems-libbsd/-/merge_requests/156.

physwkim:
Closing the loop on the upstream side: the remaining pieces have all merged, on both branches. RTEMS 7 (kernel main / libbsd 7-freebsd-14): rtems!1381, rtems!1383, libbsd !153 and !156. RTEMS 6 (kernel 6 / libbsd 6-freebsd-14): libbsd !154 and !159, with the kernel side in rtems!1439. So the kqueue path this PR enables from RTEMS 6.3 is backed on both, and the duplicate pipe()/pipe2() definitions are gone with it.

Thanks @kiwichris for the reviews and the version condition, @hjunkes for the on-target testing, and @mdavidsaver for the direction.

PR #198, opened 2026-07-27
server: contain exceptions thrown by user callbacks

physwkim:
`MonitorSetupOp::connect()` throws for a pvRequest which selects no fields, and only `SharedPV` guards it (`connectSub()`), so with `SingleSource`, `GroupSource`, or any Source written from the docs, a client sending `field(nosuchfield)` reset the whole TCP circuit and dropped its other channels. On the DESTROY_REQUEST path the throw also escapes `ServerOp::cleanup()` before `state=Dead`, so `~ServerOp()` trips its assert. The new `testmon` cases fail without this change; the first of them needs no throwing user code at all.

Callbacks reached through `acceptor_loop.dispatch()` (`onLowMark()`, `onHighMark()`, the dispatched closer) are already covered by `evbase::Pvt::doWork()` and are untouched; the `CMD_CANCEL_REQUEST` guard has no test because pvxs' own client never sends that message.

### DIFF
--- a/src/serverchan.cpp
+++ b/src/serverchan.cpp
@@ -55,8 +55,14 @@ void ServerChan::cleanup()
     }
 
     auto fn(std::move(onClose));
-    if(fn)
-        fn("");
+    if(fn) {
+        try {
+            fn("");
+        }catch(std::exception& e){
+            log_err_printf(connsetup, "Channel \"%s\" onClose() error: %s\n",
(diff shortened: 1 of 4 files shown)


mdavidsaver:
Applied with modifications as cfcdff463c7ecc521b64a3602384a7c0fe2c6c0a.  Thanks.

PR #215, opened 2026-09-11
Fix channel cache GC mark/sweep and chanByCID leak

bisegni:
Fix two issues in the client channel cache garbage collection:

  - chanByCID leak: Expired weak_ptr entries in chanByCID were never removed, causing the map to grow indefinitely and wasting CID space. Added compaction in cacheClean() to erase expired entries after sweeping chanByName.
  - GC mark/sweep logic: The cacheClean() mark/sweep correctly marks unused channels as garbage on the first pass and sweeps them on the second. cacheClear() already calls cacheClean() twice internally to ensure both phases run.

  Tests (testgc.cpp):
  - Verify channel presence in cache before sweep and absence after sweep via report()
  - Verify channel reuse by checking single channel count and growing tx bytes across consecutive gets
  - Verify Disconnect action triggers immediate sweep and disconnect callback


mdavidsaver:
I picked off part of this PR as 928d5721c121fc98b27eacc6626b5d36743278b5, and made a63627fcfd4f66ac5d1705b0fd5d96850b9846da to ensure consistence between the two `chanBy*` maps.

bisegni:
great thanks!

PR #216, opened 2026-09-12
ci: add ABICheck shadow ABI scan

napetrov:
## Summary
- add an advisory ABICheck shadow integration for libpvxs and libpvxsIoc
- build selected old/new revisions separately with side-specific public headers and source evidence
- require complete analysis and canonical JSON/Markdown reports

## Validation
- sh -n .ci-local/abicheck-diff.sh
- latest ABICheck CLI invocation smoke test
- diff check against current upstream master

ABICC remains authoritative during shadow adoption.


napetrov:
will play a bit with setup , would return back to ready once ready

Issue #214, opened 2026-09-11
Including TOOLCHAIN_PVXS.$T_A from CONFIG_PVXS_MODULE can break parallel builds


ralphlange:
Base's `CONFIG` does a generic, unconditional wildcard include of `cfg/CONFIG*` that every module build goes through.

`cfg/CONFIG_PVXS_MODULE` (installed by modules/pvxs/setup) does a plain `include` (not `-include`) of `cfg/TOOLCHAIN_PVXS.$(T_A)`, which is also generated and installed during the pvxs build.

Any module that is built in parallel and happens to include `cfg/CONFIG_PVXS_MODULE` before `cfg/TOOLCHAIN_PVXS.$(T_A)` is installed will fail its build → race condition, intermittent failure.

ralphlange:
@anjohnson: This could also be solved in the Base build system by using a staging area for `cfg` targets and only install them at the end, when the module build finished successfully.

mdavidsaver:
> Any module that is built in parallel and happens to include `cfg/CONFIG_PVXS_MODULE` before `cfg/TOOLCHAIN_PVXS.$(T_A)` is installed will fail its build → race condition, intermittent failure.

@ralphlange Have you observed a build failure like this?

I think (or maybe now thought) that `cfg/CONFIG_PVXS_MODULE` would only be included by modules which mention `PVXS` in their `RELEASE`.  Such modules declaring a build dependency should be ordered to start after PVXS has built.

I guess this issue could come up if/when multiple modules are installing to the same INSTALL_LOCATION concurrently?

mdavidsaver:
> I guess this issue could come up if/when multiple modules are installing to the same INSTALL_LOCATION concurrently?

... Which imo. seems an uncertain build process.

ralphlange:
We had a build failure last night, yes. 

I am including PVXS as a submodule inside EPICS Base, without specific ordering in the `modules/Makefile`, as it only requires the EPICS build system, which is there from the beginning. 

We're running `-j3`, so the traditional PVA modules are built in parallel. 
Always worked fine (for years), until last night, when the NT module hit the right moment.

mdavidsaver:
> I am including PVXS as a submodule inside EPICS Base

Well... topical.  https://github.com/epics-base/epics-base/issues/965

ralphlange:
What a coincidence, right?!
(That's why I created this upstream ticket instead of just ignoring it for the moment...)

[...]

PR #217, opened 2026-09-15
Enforce order of cfg/ installs

mdavidsaver:
Attempt to avoid the ordering issue shown by #214.

This sets up a three part special sequence to prepare three Makefile fragments installed in `cfg/`, prior to other directories like `src/` which depend on these fragments.

1. In `configure/`
    * Install `cfg/CONFIG_PVXS_VERSION`, which is self-contained
    * Emit the internal `configure/$(T_A)/TOOLCHAIN`.  Included through `configure/CONFIG` by all following directories.
2. In `setup/`
    * Emit and install `cfg/TOOLCHAIN_PVXS.$(T_A)`.  Due to naming, not included directly by dependent modules.
3. In new `setup2/`
    * Emit and install `cfg/CONFIG_PVXS_MODULE` and then `cfg/RULES_PVXS_MODULE`, in that order.

I think this will sequence will be proof against ordering issues from concurrent Makefile `include`.

Separating `setup/` and `setup2/` ensures that all `TOOLCHAIN_PVXS.$(T_A)` are written before `cfg/CONFIG_PVXS_MODULE`, which `includes` it.  In turn, `CONFIG_PVXS_MODULE` is installed before `RULES_PVXS_MODULE` which asserts that the CONFIG_* has already been included.

@anjohnson @simon-ess I am planning for a PVXS release soon, and would like to include this fix.  It seems sensible to me, but Makefiles are notoriously non-obvious.  So I would appreciate it if you could have a quick look.


simon-ess:
This seems broadly fine although I find it really quite troublesome to have _two_ setup directories like this.

I get that the ordering matters, but at least naïvely it seems possible to do this with a single setup directory if you move the contents of `setup` to `configure` and modify the rule for `TOOLCHAIN_PVXS.$(T_A)` to be
```make
TOOLCHAIN_PVXS.$(T_A): ../TOOLCHAIN_PVXS.target@ TOOLCHAIN
	$(EXPAND_TOOL) $(EXPAND_ARGS) $< $@
```
i.e. add TOOLCHAIN as a dependency

That is, if I test this locally it seems to produce the same output and seems to allow downstream deps to build fine...

anjohnson:
I agree with @simon-ess that including a second setup directory seems confusing, and apparently unnecessary. Please try his suggestion.

mdavidsaver:
> I find it really quite troublesome to have _two_ setup directories like this.

Me too.

> seems possible to do this with a single setup directory

I started with this, and observed a problem.  With a single directory `CONFIG_PVXS_MODULE` is installed after the _first_ (host) `TOOLCHAIN_PVXS.$(T_A)` is installed.  My usual testing situation builds for both `linux-x86_64` and `linux-x86_64-debug`, and I could see `CONFIG_PVXS_MODULE` being installed before `TOOLCHAIN_PVXS.linux-x86_64-debug`.

With this PR now, the order I see is the order I want:

```
$ grep 'Installing CFG' log
Installing CFG file ../../cfg/CONFIG_PVXS_VERSION
Installing CFG file ../../cfg/TOOLCHAIN_PVXS.linux-x86_64
Installing CFG file ../../cfg/TOOLCHAIN_PVXS.linux-x86_64-debug
Installing CFG file ../../cfg/CONFIG_PVXS_MODULE
Installing CFG file ../../cfg/RULES_PVXS_MODULE
```

anjohnson:
If you can have `configure/Makefile` create and install the `TOOLCHAIN_PVXS.$(T_A)` files (as @simon-ess suggests) and use `setup/Makefile` to install `CONFIG_PVXS_MODULE` you can achieve that same order.

It would be better if everything could be installed from `configure`, but I don't have time now to try to understand if that might be possible.

mdavidsaver:
> If you can have `configure/Makefile` create and install the `TOOLCHAIN_PVXS.$(T_A)`

At this moment I think I could.  Looking forward to inclusion of SPVA, an action of `configure/Makefile` inspects `openssl/opensslv.h` to decide if the optional TLS support can be enabled.  This per-target result is appended to `configure/$(T_A)/TOOLCHAIN`, which is `include`d by `setup/Makefile`.  As I am carrying this complexity on my development branch, I would rather minimize the difference with the release branch so that my day-to-day testing is more relevant.

> It would be better if everything could be installed from `configure`, but I don't have time now to try to understand if that might be possible.

I agree.  What I have at present has grown rather that been designed.  If/when you do have time, I would like to go through the whole process in and behind these `cfg/` fragments to see if there is a better structure.  Perhaps next month at BNL?

simon-ess:
@mdavidsaver - if that is your desired ordering, with a relatively minor change and a single directory I was able to get the same ordering as you:
```
$ grep 'Installing CFG' build.log 
Installing CFG file ../../cfg/CONFIG_PVXS_VERSION
Installing CFG file ../../cfg/TOOLCHAIN_PVXS.darwin-aarch64
Installing CFG file ../../cfg/TOOLCHAIN_PVXS.darwin-aarch64-debug
Installing CFG file ../../cfg/CONFIG_PVXS_MODULE
Installing CFG file ../../cfg/RULES_PVXS_MODULE
```
I can create a PR to this PR if you like.

simon-ess:
> > I find it really quite troublesome to have _two_ setup directories like this.
> 
> Me too.
> 
> > seems possible to do this with a single setup directory
> 
> I started with this, and observed a problem. With a single directory `CONFIG_PVXS_MODULE` is installed after the _first_ (host) `TOOLCHAIN_PVXS.$(T_A)` is installed. My usual testing situation builds for both `linux-x86_64` and `linux-x86_64-debug`, and I could see `CONFIG_PVXS_MODULE` being installed before `TOOLCHAIN_PVXS.linux-x86_64-debug`.
> 
> With this PR now, the order I see is the order I want:
> 
> ```
> $ grep 'Installing CFG' log
> Installing CFG file ../../cfg/CONFIG_PVXS_VERSION
> Installing CFG file ../../cfg/TOOLCHAIN_PVXS.linux-x86_64
> Installing CFG file ../../cfg/TOOLCHAIN_PVXS.linux-x86_64-debug
> Installing CFG file ../../cfg/CONFIG_PVXS_MODULE
> Installing CFG file ../../cfg/RULES_PVXS_MODULE
> ```

Actually, I'm a bit confused by what you say versus what is written here. You say you want CONFIG_PVXS_MODULE installed _before_ the toolchains, but the grep block shows the opposite order.

mdavidsaver:
> You say you want CONFIG_PVXS_MODULE installed _before_ the toolchains

This is what I saw during my initial attempt, before adding `setup2/`.  This ordering is not correct.

> the grep block shows the opposite order.

The order shown in this block is what I think is necessary, and what I now see with this PR.

mdavidsaver:
> relatively minor change and a single directory
> I can create a PR to this PR if you like.

Please do.  I like minor changes to Makefiles.

simon-ess:
Done. https://github.com/mdavidsaver/pvxs-dev/pull/2

mdavidsaver:
I think there is agreement that this PR will accomplish the desired goal.  None of us like how this is done, but I will merge this PR so that it can be included in the imminent 1.5.3 release.  I would like to see this discussion continue towards a less horrible solution.

PR #219, opened 2026-09-17
GHA: add abicheck

mdavidsaver:
Add GHA usage of abicheck.

@napetrov Is there a way to define additional C preprocessor macros when `abicheck dump` processes headers?  This library has an "expert" API which is normally hidden.  User code should define `PVXS_ENABLE_EXPERT_API` to opt-in.  I would like these added definitions to be considered public by abicheck.

I first looked for a `abicheck dump -D...` argument to parallel the usual `-I`.  I then tried `--gcc-options "-DPVXS_ENABLE_EXPERT_API"`, but saw an arg. parsing error.


napetrov:
Thanks for trying this. You are right that `-D...` is the natural CLI syntax to expect here.

At the moment, `abicheck dump` takes the stable header-compilation context from `.abicheck.yml`, so `--gcc-options` is not accepted. The current way to enable the PVXS Expert API is:

```yaml
# .ci-local/abicheck.yml
compile:
  defines:
    - PVXS_ENABLE_EXPERT_API
```

Then use the same configuration for both snapshots:

```sh
abicheck dump \
  lib/linux-*/libpvxs.so.* \
  --config .ci-local/abicheck.yml \
  -I "${HOME}/.cache/base-7.0/include" \
  -I "${HOME}/.cache/base-7.0/include/os/Linux" \
  -I "${HOME}/.cache/base-7.0/include/compiler/gcc" \
  -H include/ \
  -o /tmp/pvxs-CUR.json
```

and likewise for `pvxs-PREV.json`.

This does not change how PVXS is built. It only defines `PVXS_ENABLE_EXPERT_API` while abicheck parses the installed public headers, modeling a consumer that has opted into the Expert API. `PVXS_API_BUILDING` should not be defined because that is private to building the library itself.

Our experimental integration currently omitted this macro, so it covered only the ordinary consumer-visible API. I will update it to include the Expert declarations as requested.

I also agree that requiring a config file for this one-off case is not sufficiently intuitive. I will treat the missing dedicated, repeatable `-D`/`--define` option as an abicheck CLI gap. The intention is to add a narrow macro-definition option rather than restore unrestricted compiler-option forwarding, while keeping `.abicheck.yml` as the recommended form for reproducible CI and baseline generation.

I used `-o` above intentionally. Redirecting stdout to JSON is supported and useful for pipelines, but `-o`/`--output` is preferable when persisting a snapshot because abicheck owns the file write, performs atomic replacement, provides clearer write-error handling, and can support snapshot compression.

